PR #2, opened 2021-06-04
Doc cleanup and prettifying

lebovic:
Adds:
- Pretty printed job status updates
- PyCharm gitignore

Modifies:
- Readme/doc ordering and imports
- Function signature for tool arguments

Also: we'll need to set guidelines for PRs soon, since this is open source now.

### DIFF
--- a/toolchest_client/_internal_utils.py
+++ b/toolchest_client/_internal_utils.py
@@ -9,7 +9,7 @@ This module provides internal utility functions for running Toolchest tools.
 def _validate_tool_kwargs(**kwargs):
     """Validates kwargs passed to (specific) tool functions."""
 
-    # Defaults for query parameters input_name and output_name are provided
+    # Default values for query parameters input_name and output_name are provided
     # by default in tool function; they should not be provided by the user.
     if "input_name" in kwargs:
         raise ValueError("input_name should not be manually specified")
--- a/toolchest_client/tools.py
+++ b/toolchest_client/tools.py
@@ -24,7 +24,7 @@ def run_tool(tool, version, **kwargs):
     q = Query()
     q.run_query(tool, version, **kwargs)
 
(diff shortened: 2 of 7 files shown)


toolchest_client/_internal_utils.py
```diff
@@ -9,7 +9,7 @@
 def _validate_tool_kwargs(**kwargs):
     """Validates kwargs passed to (specific) tool functions."""
 
-    # Defaults for query parameters input_name and output_name are provided
+    # Default values for query parameters input_name and output_name are provided
     # by default in tool function; they should not be provided by the user.
```
lebovic, line 13: What do you mean here by "they should not be provided by the user?"

It seems reasonable that someone would run this like:
```
>>> toolchest.cutadapt(
... tool_args="",
... input_path="./temp/short.fastq",
... output_path="./output.fastq"
... )
```

toolchest_client/tools.py
```diff
@@ -40,9 +40,9 @@ def cutadapt(cutadapt_args, **kwargs):
 
     Usage::
 
-        >>> import toolchest_client as tc
-        >>> tc.cutadapt(
-        ...     "-a AACCGGTT",
+        >>> import toolchest_client as toolchest
+        >>> toolchest.cutadapt(
+        ...     tool_args="-a AACCGGTT",
```
lebovic, line 45: Specifically calling out this change – what are you thoughts?

bcai2 (approved):
lgtm!

toolchest_client/_internal_utils.py
lebovic: @bcai2 can you still respond to this?

toolchest_client/tools.py
bcai2: I'm okay with this. It's less concise but more explicit.

PR #4, opened 2021-06-29
Configure Poetry for dependency management

lebovic:
Transitions to Poetry for dependency management and interfacing with PyPI.

Adds:
- Poetry! (https://python-poetry.org/)

Modifies:
- How to add packages (https://python-poetry.org/docs/cli#add)
- How to deploy to PyPI (this will be in CircleCI and automatic soon)

Removes:
- Twine


bcai2 (approved):
LGTM. (I've also heard of [Dependabot](https://dependabot.com/), if that seems like something that would be handy.)

lebovic:
> LGTM. (I've also heard of [Dependabot](https://dependabot.com/), if that seems like something that would be handy.)

GitHub bought Dependabot, so this is now enabled by default :)

PR #5, opened 2021-06-29
Add basic CircleCI config

lebovic:
Adds a basic CircleCI config. We'll also want to add lint and tests to CircleCI later.

Adds:
- Automatic deployment of builds to PyPI. `staging` deploys to TestPyPI (tested), and `main` deploys to normal PyPI (untested).


bcai2 (approved):
Nice, LGTM. We'll probably want to update the version # once we merge the `bowtie2` branch, which has all the nice features we want.

PR #7, opened 2021-07-02
Add test pipeline segment

lebovic:
Adds a `test` pipeline segment.

Adds:
- `test` pipeline segment to allow demonstrating/trying out the interface without a specific tool to distract.
- `.env` handling with `python-dotenv`

### DIFF
--- a/toolchest_client/query.py
+++ b/toolchest_client/query.py
@@ -25,7 +25,7 @@ class Query():
     """
 
     # Base URLs used by the server API.
-    BASE_URL = "https://api.toolche.st"
+    BASE_URL = os.environ.get("BASE_URL", "https://api.toolche.st")
     PIPELINE_ROUTE = "/pipeline-segment-instances"
     PIPELINE_URL = BASE_URL + PIPELINE_ROUTE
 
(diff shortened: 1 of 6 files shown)


toolchest_client/query.py
```diff
@@ -25,7 +25,7 @@ class Query():
     """
 
     # Base URLs used by the server API.
-    BASE_URL = "https://api.toolche.st"
+    BASE_URL = os.environ.get("BASE_URL", "https://api.toolche.st")
```
lebovic, line 28: Note that this isn't setting a `DEPLOY_ENVIRONMENT` environment variable.

To set, you can create a `.env` file in the root of the project.

toolchest_client/tools.py
```diff
@@ -120,3 +122,31 @@ def kraken2(tool_args="", **kwargs):
         output_name="output.txt",
         **kwargs
     )
+
+
+def test(tool_args="", **kwargs):
+    """Run a test pipeline segment via Toolchest. A plain text file containing 'step one' is returned."
```
bcai2, line 128: Is the last quote a typo? Otherwise, everything else looks good.

bcai2 (approved):
LGTM, see comment about a typo

toolchest_client/tools.py
lebovic: Whoops, yep. It's `success` now.

PR #10, opened 2021-07-09
Change tool argument order

bcai2:
Changes:

* moved `tool_args` to the front of the declared arguments for every tool function argument (for consistency + ease of usability when porting to R)


lebovic (approved):
👍

PR #13, opened 2021-07-21
Override built-in print statement to flush after every print

lebovic:
Adds:
- Support for consistent buffer flushing. This allows consistent output when loaded in R via `reticulate`


bcai2 (approved):
LGTM

PR #15, opened 2021-07-22
Add Unicycler and named inputs

lebovic:
Adds:
- Support for Unicycler 0.4.9
- Support for multiple named inputs


lebovic (commented):
Skipping formal review, as we reviewed together.

PR #18, opened 2021-08-11
Added basic tool_args sanitization

bcai2:
Added basic sanitization to custom `tool_args` provided by the user, in the form of a whitelist of Toolchest-supported tags in `arg_whitelist.py`. 

TODO: remove non-escaped command-line characters that would be interpreted as non-text in bash. (This should be easy to do with a properly formatted regex.)


lebovic (approved):
Looks good! A feature like this really highlights our lack of unit tests – perhaps we should up the priority of adding them.

PR #20, opened 2021-08-17
STAR base support

bcai2:
Adds base support for STAR.

Output is currently restricted to just the aligned reads (the `.sam` file). The compression step can be implemented in another PR.

Database selection is currently disabled and defaults to GRCh38.


toolchest_client/tools/api.py
```diff
@@ -124,17 +151,17 @@ def test(inputs, output_path, tool_args=""):
 
     instance = Test(
         tool_args=tool_args,
-        output_name='output.txt',
+        output_name='Aligned.out.sam',
```
lebovic, line 154: Was this intentional?

toolchest_client/tools/api.py
```diff
@@ -124,17 +151,17 @@ def test(inputs, output_path, tool_args=""):
...
     )
     instance.run()
 
 
 def unicycler(output_path, read_one=None, read_two=None, long_reads=None, tool_args=""):
-    """Runs Bowtie 2 (for alignment) via Toolchest.
+    """Runs Unicycler (for alignment) via Toolchest.
 
-    :param tool_args: (optional) Additional arguments to be passed to Bowtie 2.
+    :param tool_args: (optional) Additional arguments to be passed to Unicycler.
```
lebovic, line 164: Whoops, thanks

lebovic (approved):
LGTM at a glance

toolchest_client/tools/api.py
```diff
@@ -105,6 +105,33 @@ def kraken2(inputs, output_path, tool_args=""):
...
+        ...     tool_args="--mode bold",
+        ...     output_path="scratch"
+        ... )
+
+    """
+
+    instance = STAR(
+        tool_args=tool_args,
+        output_name='output.zip',
+        inputs=[read_one, read_two],
```
lebovic, line 131: Won't this require an `input_prefix_mapping`? It might break the pattern currently employed, which assumes that each input corresponds to one tool arg

toolchest_client/tools/api.py
bcai2: Yes. For now, we don't have control over the output names of STAR, and before we implement multiple output files + compressing multiple files into one outputted zip, we should just output the `.sam` file.

lebovic: This is the `Test` instance, not `STAR`

lebovic: Should have specified what I was commenting on, whoops

toolchest_client/tools/api.py
bcai2: It's hard to say since STAR's syntax requires specifying multiple input filepaths under the same tag, so we'd need to expand our current implementation (unless I'm misunderstanding something).

lebovic: Huh, actually looks like it'll be fine. On the worker node, inputs with no prefix are translated to an arg with a value looking like f"{path} " (note the space). So default handling should work with STAR.

toolchest_client/tools/api.py
bcai2: My bad -- should be fixed now

PR #21, opened 2021-08-18
Add updated Kraken2 options

bcai2:
Adds database selection for `kraken2`. Defaults to `standard` (the standard Kraken2 DB) for backwards compatibility.


lebovic (approved):
LGTM

PR #24, opened 2021-08-20
Fix support for Unicycler

lebovic:
Adds:
- Explicit concept of jobs in Tools to distinguish between Tools which have multiple input files and parallel runs which have multiple split input files.

Fixes support for Unicycler (non-parallel) in v0.7.x.


lebovic:
@bcai2 merging as a quick fix

PR #27, opened 2021-08-24
Fix/error propagation loop

bcai2:
Fixes loop upon encountering an in-job error between `_get_job_status()` and `_raise_for_failed_job()` (now `_raise_for_failed_response()`) in `query.py`.

(I'm tabling suppressing error messages upon encountering an error for another PR.)

### DIFF
--- a/toolchest_client/query.py
+++ b/toolchest_client/query.py
@@ -319,6 +319,9 @@ class Query():
     def _mark_as_failed_at_exit(self):
         """Upon exit, marks job as failed if it has started but is not marked as completed/failed."""
 
+        print("at atexit")
+        return
+
         # TODO: look at putting this function inside the __init__?
         # otherwise, each Query instance persists until exit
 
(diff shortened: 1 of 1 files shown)


toolchest_client/query.py
```diff
@@ -185,51 +185,49 @@ def _upload(self, input_file_paths, input_prefix_mapping):
                 upload_response.raise_for_status()
             except HTTPError as e:
                 # todo: this isn't propagating as a failure
```
lebovic, line 187: Is this todo still true? I added it – and forgot to remove it – when testing STAR parallelization. I see that you added a `force_raise=True`.

lebovic (approved):
👍

toolchest_client/query.py
lebovic: Update: looks like this was just due to a default value change.

bcai2: I'm not sure if there's still work to be done here. Could you share what inputs invoked the non-propagated failure?

PR #35, opened 2021-08-27
Basic unit tests for files module

bcai2:
Adds basic unit tests for all functions in `toolchest_client/files` apart from `merge_sam_files()`.


toolchest_client/files/split.py
```diff
@@ -26,7 +26,7 @@ def open_new_output_file(
     if not os.path.exists(working_directory):
         os.mkdir(working_directory)
     current_output_file_path = f"{working_directory}/{filename_prefix}_{current_split_number}_{input_basename}"
-    return current_output_file_path, open(current_output_file_path, "w")
+    return current_output_file_path, open(current_output_file_path, "w", newline="\n")
```
bcai2, line 29: This is to make tests cooperate on Windows. `open()` on Windows will auto-write Windows-style `\r\n` instead of `\n` for line endings.

toolchest_client/files/tests/test_general.py
```diff
@@ -0,0 +1,50 @@
...
+    for file in file_paths:
+        open(file, "w")
+    file_paths = sorted([os.path.normpath(x) for x in file_paths])
+
+    assert sorted([os.path.normpath(x) for x in files_in_path(tmp_dir)]) == file_paths
+
+    for file in file_paths:
+        os.remove(file)
+    os.rmdir(sub_dir)
+    os.rmdir(tmp_dir)
```
bcai2, line 36: Not sure if there's a better way to do this; if the `assert` fails, then the `tmp` files will remain in the `files/tests` directory once tests end.

lebovic: You could add:
a) a pytest fixture that runs before and after each test
b) a try-catch-finally

(a) would be pretty non-generalizable, (b) seems more apt

toolchest_client/files/tests/test_general.py
```diff
@@ -0,0 +1,50 @@
...
+def test_files_in_path():
+    tmp_dir = f"{THIS_FILE_PATH}/tmp"
+    tmp1 = f"{tmp_dir}/tmp1"
+    tmp2 = f"{tmp_dir}/tmp2"
+    sub_dir = f"{tmp_dir}/sub_dir"
+    tmp3 = f"{sub_dir}/tmp3"
+    file_paths = [tmp1, tmp2, tmp3]
+    for dir in [tmp_dir, sub_dir]:
+        if not os.path.exists(dir):
+            os.mkdir(dir)
```
lebovic, line 26: Could also: `os.makedirs(dir, exist_ok=True)`

toolchest_client/files/tests/test_general.py
```diff
@@ -0,0 +1,50 @@
...
+    for file in file_paths:
+        open(file, "w")
```
lebovic, line 28: Is the intent here to just create a blank new file?

lebovic (approved):
Looks good, a few questions

toolchest_client/files/tests/test_general.py
bcai2: Yes, just for testing.

lebovic: In that case you can chain a `.close()` to the `open` call to immediately close the file

toolchest_client/files/tests/test_general.py
bcai2: A follow-up to the try-catch-finally comment (since I can't reply to it directly, for some reason): looks like putting the `assert` in the `try` statement suppresses the `AssertionError` entirely, so I'm just going to leave it like this

PR #36, opened 2021-09-18
Key validation restructuring

bcai2:
Adds `_validate_key()` to `auth.py` and moves the key validation check from `Query.run_query()` to `Tool.run()`, where it occurs before any jobs are spawned.

Note that any invalid key checks can take as long as a minute. The key validity is determined to be invalid by the API, which hangs and eventually times out upon encountering an invalid key. We'll need to make a change to the API to speed this up as well.


lebovic (approved):
👍

PR #37, opened 2021-09-20
v0.7.14

lebovic:


bcai2:
I'm thinking about merging this in now just to make testing more seamless on the R side, since that solely depends on the `main` release of the Python client.

PR #39, opened 2021-10-15
Enable paired reads for `kraken2`

bcai2:
Adds the option to use paired-read inputs for `kraken2`, via the `read_one` and `read_two` arguments (or a list of two paths via `inputs`).

Adds/removes `--paired` to `tool_args` as necessary.

### DIFF
--- a/tests/test_tools.py
+++ b/tests/test_tools.py
@@ -59,4 +59,4 @@ def test_kraken2_paired_end():
         output_path=output_file_path,
     )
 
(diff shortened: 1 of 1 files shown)


bcai2:
@lebovic Can you upload paired-end sample data? All the data I can find on SRA is far too large for practical testing. 

Let me know when that's done so I can merge this in. I've added a placeholder for a paired-end Kraken2 test.

PR #44, opened 2021-11-02
Add shi7

lebovic:
Adds:
- `shi7` tool
- `shi7` single-end integration test
- `shi7` paired-end integration test, but it's non-deterministic so it just confirms it's not the same as single-end

Inputs are artificially constrained at 10 for now.


toolchest_client/tools/api.py
```diff
@@ -109,8 +109,8 @@ def kraken2(output_path, inputs=[], database_name="standard", database_version="
 
     if read_one:
         inputs = [read_one]
-    if read_two:
-        inputs.append(read_two)
+        if read_two:
+            inputs.append(read_two)
```
lebovic, line 113: Not related to shi7, but formerly the R2 (only) could be added in paired-end mode.

toolchest_client/tools/tool.py
```diff
@@ -73,12 +73,11 @@ def _validate_args(self):
...
         if self.output_path is None:
             raise ValueError("No output path provided.")
-        try:
-            with open(self.output_path, "a") as _:
-                pass
-        except OSError:
+        if not os.access(
+                os.path.dirname(self.output_path),
+                os.W_OK | os.X_OK,
+        ):
```
lebovic, line 79: Prevents creating an empty file, such that:
```
toolchest.shi7(
    inputs="./directory/",
    output_path="./directory/,
)
```

works.

toolchest_client/tools/api.py
bcai2: We might want to add a note to the docstring about this, but I think that the docs/docstrings are getting unwieldly and may in need of restructuring soon, so I'm fine with specifying this later.

toolchest_client/tools/api.py
```diff
@@ -135,6 +135,32 @@ def kraken2(output_path, inputs=[], database_name="standard", database_version="
     instance.run()
 
 
+def shi7(inputs, output_path, tool_args=""):
+    """Runs shi7 via Toolchest.."
```
bcai2, line 139: typo at the end of this line?

bcai2 (approved):
LGTM; found a typo(?) and made a note about docs, but ready to merge when you are :+1:

toolchest_client/tools/api.py
lebovic: Whoops, yep.

PR #47, opened 2021-11-09
Adds parallelization for shi7 and shogun (BURST)

lebovic:
Adds:
- Parallelization for `shi7`
- Parallelization for `shogun` (BURST)

Note that this is not fully validated and should not yet be used in production.


toolchest_client/files/split.py
```diff
@@ -68,11 +72,49 @@ def split_file_by_lines(input_file_path, num_lines_in_group=4, max_bytes=4.5 * 1
...
+
+    grouped_input_file_paths = {}
+    for input_file_path in input_file_paths:
+        no_read_file_path = re.sub(r'R\d', '', input_file_path)
+        if grouped_input_file_paths.get(no_read_file_path):
+            grouped_input_file_paths[no_read_file_path].append(input_file_path)
+        else:
+            grouped_input_file_paths[no_read_file_path] = [input_file_path]
+
+    for _, [read_one_file_path, read_two_file_path] in grouped_input_file_paths.items():
```
bcai2, line 103: If no pair is found (i.e., some reads are standalone single-end), is the resulting behavior intended?

bcai2 (approved):
LGTM, there is one comment to address

toolchest_client/files/split.py
lebovic: If no pair is found, we'd end up with something like:

```
[yields pairs until the non-matched file is found]
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ValueError: not enough values to unpack (expected 2, got 1)
```

It's not ideal; ideally we should be confirming that there are an even number of matching R1s and R2s. Will resolve.

PR #49, opened 2021-11-10
S3 URI inputs

bcai2:
Adds support for inputs that are provided as AWS S3 URIs.

Parallelization is disabled for any job that uses an S3 URI input, since this would require redownloading the file from S3, splicing locally, and reuploading the spliced inputs to S3.

Note: Uploading is skipped since the input file will already be in S3. However, the status is still updated to `TRANSFERRING_FROM_CLIENT` and `TRANSFERRING_TO_CLIENT`, like the status path for a locally uploaded file.


toolchest_client/tools/tool.py
```diff
@@ -264,24 +274,26 @@ def _generate_jobs(self, should_run_in_parallel):
...
+                    yield input_file_path_pair, PARALLEL_FILE_IS_IN_S3
 
         else:
             # Make sure we're below plan/multi-part limit for non-splittable files
-            for file_path in self.input_files:
-                check_file_size(file_path, max_size_bytes=FOUR_POINT_FIVE_GIGABYTES)
+            for file_path, file_is_in_s3 in zip(self.input_files, self.inputs_are_in_s3):
+                if not file_is_in_s3:
+                    check_file_size(file_path, max_size_bytes=FOUR_POINT_FIVE_GIGABYTES)
+                # If the file is already in S3, no need to check max size.
```
lebovic, line 292: Should we guard against large input files from S3? Could cause us to exceed resource limits (e.g. a 200 GB input file causing an out of memory issue for Kraken 2)

lebovic (approved):

toolchest_client/api/query.py
```diff
@@ -152,39 +155,53 @@ def _add_input_file(self, input_file_path, input_prefix):
...
-        for file_path in input_file_paths:
-            print(f"Uploading {file_path}")
-            upload_url = self._add_input_file(
-                input_file_path=file_path,
-                input_prefix=input_prefix_mapping.get(file_path)
-            )
+        for file_path, file_is_in_s3 in zip(input_file_paths, inputs_are_in_s3):
+            # If the file is already in S3, there is no need to upload.
+            if file_is_in_s3:
+                # Registers the file in the internal DB.
```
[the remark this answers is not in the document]
bcai2: There's a whole host of things (status logging, output messages about "uploading") that are currently being treated the same way as a file uploaded by the user. IMO, it's best to group that together in a separate PR.

toolchest_client/tools/tool.py
bcai2: Good point -- we can merge this in for now and implement later.

lebovic: To clarify: is there any guard for the 5 GB limit on S3 files right now?

bcai2: Not right now. I'll update this with a quick fix PR later today.

toolchest_client/api/query.py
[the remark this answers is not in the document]
lebovic: Just saw this now. I disagree – any UI changes that are tied to functionality change should go through at the same time as the functionality change.

PR #51, opened 2021-11-21
Add Zoey suggested changes

lebovic:
Modifies:
- No longer requires an empty output directory.
- A check for overwriting files was moved to a preflight check.

Still needs:
- Better docs.

### DIFF
--- a/toolchest_client/api/auth.py
+++ b/toolchest_client/api/auth.py
@@ -48,7 +48,7 @@ def set_key(key):
         os.environ["TOOLCHEST_KEY"] = key
 
 
-def _validate_key():
+def validate_key():
     """Validates Toolchest API key, retrieved from get_key()."""
 
     BASE_URL = os.environ.get("BASE_URL", "https://api.toolche.st")
--- a/toolchest_client/files/unpack.py
+++ b/toolchest_client/files/unpack.py
@@ -10,7 +10,7 @@ class OutputType(Enum):
     FLAT_TEXT = ".txt"
 
 
-def link_dst_unlink_src(src, dst):
+def move_from_child_to_parent_dir(src, dst):
     """Hard links files within src to dst, and then unlinks the src. dst must be a parent of src."""
     src_path = Path(src)
     dst_path = Path(dst)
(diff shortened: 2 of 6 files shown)


toolchest_client/api/auth.py
```diff
@@ -48,7 +48,7 @@ def set_key(key):
         os.environ["TOOLCHEST_KEY"] = key
 
 
-def _validate_key():
+def validate_key():
```
lebovic, line 51: Renamed, because it's not being used as a private method

toolchest_client/files/unpack.py
```diff
@@ -10,7 +10,7 @@ class OutputType(Enum):
     FLAT_TEXT = ".txt"
 
 
-def link_dst_unlink_src(src, dst):
+def move_from_child_to_parent_dir(src, dst):
```
lebovic, line 13: Renamed for clarity

toolchest_client/files/unpack.py
```diff
@@ -43,26 +43,17 @@ def unpack_files(file_path_to_unpack, output_type):
...
-        # The underlying shutil code does some, but it's limited
         unpacked_outputs_final_destination = os.path.dirname(file_path_to_unpack)
-        # Make sure the directory is empty except for the GZ_TAR
-        visible_files_in_directory = [
-            f for f in os.listdir(unpacked_outputs_final_destination) if not f.startswith('.')
-        ]
-        if len(visible_files_in_directory) > 1:
-            raise ValueError(
-                f"{unpacked_outputs_final_destination} is not empty. Please use an empty directory for output!"
-            )
```
lebovic, line 55: Instead of throwing an error while unpacking, we now print a warning in the beginning if there are any files which might conflict / be overwritten

toolchest_client/tools/tool.py
```diff
@@ -148,11 +148,13 @@ def _merge_outputs(self, output_file_paths):
...
-        Confirms that – at a very basic level – output was created.
-        Tool subclasses can have more specific implementations.
-        """
+    def _preflight(self):
+        """Generic preflight check. Tools can have more specific implementations."""
+        # Validate Toolchest auth key.
+        validate_key()
+
+    def _postflight(self):
+        """Generic postflight check. Tools can have more specific implementations."""
```
lebovic, line 157: Created `preflight` and `postflight` methods, which can be overwritten or extended by specific tool implementations. See above for the Kraken 2 implementation.

This will make better validation easier in the future for other use cases as well.

toolchest_client/tools/kraken2.py
```diff
@@ -32,7 +32,26 @@ def __init__(self, tool_args, output_name, inputs, output_path,
             output_is_directory=True,
         )
 
-    def _sanity_output_check(self):
+    def _preflight(self):
```
bcai2, line 35: At some point, can you add a comment line about what specifically the Kraken 2 preflight is doing? 

(It's evident right now, but this will help for looking back at this code later.)

bcai2 (commented):
Sorry for the late reply -- LGTM, though could use an extra comment. Thanks for implementing this!

toolchest_client/tools/kraken2.py
lebovic: We can! I've added this blurb to the ongoing `fix/update-docs` branch:

```
    def _preflight(self):
        """
        Kraken 2 specific preflight.
        
        Checks:
        - that `output_path` either: a) exists and isn't a file, or b) can and does create the output path.
        - if the files that Kraken 2 will write already exist. If they do, prints a warning.
        """
        super()._preflight()
```

That said, I'm actually in favor of not adding a comment. I could grab the above context from the above code more easily than by reading a comment, although I'm likely biased.

PR #52, opened 2021-11-23
MEGAHIT support

bcai2:
Adds:
- support for `megahit`, via `toolchest.megahit()`.

Of note:
- `megahit` processes inputs via 4 different tags (`-1`, `-2`, `--12`, `-r`) that can each represent an arbitrarily large number of input files (comma-delineated in a single string). 
- Input files can be passed in as an individual path (`'/path/to/file'`) or as a list of paths (`['/path/to/file1', '/path/to/file2']`). When inputs are given in a list, the lists are flattened into one large `inputs` list, and each input file is registered in the API with its associated input tag (`-1`, `-2`, `--12`, `-r`). (If a tag has multiple inputs, the tag argument is subsequently processed by the worker node.)
- Though `megahit` can take in an arbitrarily large number of input files, the input file limit for megahit is set at 10.


toolchest_client/tools/api.py
```diff
@@ -135,7 +135,7 @@ def kraken2(output_path, inputs=[], database_name="standard", database_version="
         >>> toolchest.kraken2(
         ...     tool_args="",
         ...     inputs="./path/to/input.fastq",
-        ...     output_path="./path/to/output.txt",
+        ...     output_path="./path/to/output.tar.gz",
```
lebovic, line 138: This should be modified, but it shouldn't be a .tar.gz. It should be a directory, e.g. `./path/to/output`.

The compression and decompression happens in the background. Try using the Kraken 2 tool to see how it works.

Overall, I think it's time to revamp our documentation.

toolchest_client/tools/api.py
```diff
@@ -168,6 +168,61 @@ def kraken2(output_path, inputs=[], database_name="standard", database_version="
...
+    input_list = []  # list of all inputs
+    input_prefix_mapping = {}  # map of each input to its respective tag
+    for tag, param in tag_to_param_map.items():
+        if isinstance(param, list):
+            for input_file in param:
+                input_list.append(input_file)
+                input_prefix_mapping[input_file] = tag
+        elif isinstance(param, str):
+            input_list.append(param)
+            input_prefix_mapping[param] = tag
```
lebovic, line 214: This is more a question than a comment.

Our code in api.py file is getting more and more complex. I wonder if there's a better place for logic like this?

toolchest_client/tools/api.py
```diff
@@ -222,7 +277,7 @@ def shogun_align(inputs, output_path, database_name="shogun_standard", database_
         inputs=inputs,
         output_path=output_path,
         database_name=database_name,
-        database_version=database_version
+        database_version=database_version,
```
lebovic, line 280: Not related to this PR, but it's time to add a linter to this repo.

toolchest_client/tools/megahit.py
```diff
@@ -0,0 +1,26 @@
...
+            tool_name="megahit",
+            tool_version="1.2.9",  # todo: allow version to be set by the user
+            tool_args=tool_args,
+            output_name=output_name,
+            output_path=output_path,
+            inputs=inputs,
+            input_prefix_mapping=input_prefix_mapping,
+            min_inputs=1,
+            max_inputs=10,  # todo: make this unlimited?
+        )
```
lebovic, line 26: Nit: missing newline

lebovic (approved):

lebovic:
@bcai2 merging prior to your re-review, because you wrote it :)

PR #55, opened 2021-11-23
Add pytest + CircleCI parallelism

lebovic:
CircleCI has three default ways that you can split `pytest` tests for parallelization: by file size (literal file size), file/class name (literally splitting by file/class name), or by historical timings for running the tests.

The ultimate goal is to split by historical timings. In order to do that, we first need to have historical data. Thus, to start, we'll split by file size.

Another fun quirk is that the CircleCI executor will fail if the `parallelism` count is higher than the number of "splits". For example, if there is a parallelization factor of 10 but only 3 files with tests, it will fail. For that reason, we have to keep the parallelization count lower than the number of splits. Similarly, it seems to only split by file rather than individual tests; this is an issue because we have many multiple-minute tests per file.

This isn't the best split or parallelization, but it's a step in the right direction.

Adds:
- Parallelization for CircleCI / `pytest`, with an initial parallelization of 2.

Modifies:
- The file structure for integration tests. Refactors the large `test_tools.py` file to multiple files, and adds a `pytest` `conftest.py`.


.circleci/config.yml
```diff
@@ -33,7 +33,7 @@ jobs:
   integration-tests:
     executor: python/default
     working_directory: ~/repo
-    parallelism: 10
+    parallelism: 2
```
lebovic, line 36: This is a diff from 10 -> 2, simply because I tried to add parallelism in the staging branch. I incorrectly assumed this would be a couple line change to add parallelization :)

.circleci/config.yml
```diff
@@ -54,7 +54,7 @@ jobs:
             cd ~/repo
             export DEPLOY_ENVIRONMENT=$([[ $CIRCLE_BRANCH = main ]] && echo production || echo staging)
             shopt -s globstar
-            TESTFILES=$(circleci tests glob tests/**/test*.py | circleci tests split --split-by=timings)
+            TESTFILES=$(circleci tests glob tests/**/test*.py | circleci tests split --split-by=filesize)
```
lebovic, line 57: Same here as above

tests/conftest.py
```diff
@@ -0,0 +1,11 @@
+import os
+
+
+def pytest_configure(config):
```
lebovic, line 4: This runs before all tests in the directory

lebovic:
@bcai2 I'm going to merge this prior to review – it's largely just a refactor, and is devops focused – but I would still like your review.

PR #56, opened 2021-12-08
Limit on S3 URI input size

bcai2:
Adds a check on the size of an input provided as an S3 URI, as a safeguard. The limit is the same as the size of a user-uploaded file (4.5 GB).

Note: This limit will be removed once multipart uploads are implemented for client-uploaded inputs (so that the behavior of the limts for S3 URI inputs matches), though we may want to increase the size of the limit rather than remove it altogether.

### DIFF
--- a/toolchest_client/files/general.py
+++ b/toolchest_client/files/general.py
@@ -8,7 +8,9 @@ General file handling functions.
 import shutil
 import os
 
-from .s3 import assert_accessible_s3
+import boto3
+
+from .s3 import assert_accessible_s3, get_params_from_s3_uri
 
 
 def assert_exists(path, must_be_file=False, must_be_directory=False):
@@ -30,7 +32,7 @@ def assert_exists(path, must_be_file=False, must_be_directory=False):
         raise ValueError(f"Directory entry at {path} is not a directory")
 
 
-def check_file_size(file_path, max_size_bytes=None):
+def check_file_size(file_path, max_size_bytes=None, is_s3_uri=False):
     """Raises an error if the file is above the non-multipart upload limit for S3 (5GB)
 
     :param file_path: A path to a file.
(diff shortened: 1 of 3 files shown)


toolchest_client/tools/tool.py
```diff
@@ -289,9 +289,8 @@ def _generate_jobs(self, should_run_in_parallel):
         else:
             # Make sure we're below plan/multi-part limit for non-splittable files
             for file_path, file_is_in_s3 in zip(self.input_files, self.inputs_are_in_s3):
-                if not file_is_in_s3:
-                    check_file_size(file_path, max_size_bytes=FOUR_POINT_FIVE_GIGABYTES)
-                # If the file is already in S3, no need to check max size.
+                check_file_size(file_path, max_size_bytes=FOUR_POINT_FIVE_GIGABYTES, is_s3_uri=file_is_in_s3)
```
lebovic, line 292: Nit: `is_s3_uri` and `file_is_in_s3` represent the same thing but have different names

toolchest_client/tools/tool.py
```diff
@@ -289,9 +289,8 @@ def _generate_jobs(self, should_run_in_parallel):
...
+                # If the file is already in S3, the size is checked in check_file_size as well.
```
lebovic, line 293: Nit: spacing

toolchest_client/files/general.py
```diff
@@ -30,19 +32,31 @@ def assert_exists(path, must_be_file=False, must_be_directory=False):
...
+        file_size_bytes = os.stat(file_path).st_size
+    else:
+        # Get file size S3 metadata, via boto3.
+        s3_file_params = get_params_from_s3_uri(file_path)
+        s3_client = boto3.client("s3")
+        response = s3_client.head_object(
+            Bucket=s3_file_params["bucket"],
+            Key=s3_file_params["key"],
+        )
+        file_size_bytes = response["ContentLength"]
```
lebovic, line 54: This check be better suited for the API, where we actually can enforce data sizes (barring a race condition exploit where the file is replaced on S3 after a size check).

I'm fine with this existing here, but worth noting.

lebovic (approved):

lebovic:
Also, the multipart upload limit isn't restricting downloads from S3. This is our arbitrary limit to make sure S3 file sizes match what we're expecting from the client, until we remove that assumption from the infra side.

PR #22, opened 2021-08-18
Add improved time display

bcai2:
Changes outputted time elapsed during execution to `(h:mm:ss`) instead of `(#s)`.

PR #57, opened 2021-12-09
Polish tool_arg handling, add more STAR args

lebovic:
Adds:
- More STAR args
- Add multiple levels of tool_arg handling (whitelist, dangerlist, blacklist)
- Error on unknown or blacklisted args
- Reduce complexity (validation and parallelization for now) if a dangerous argument is passed

Requires:
- https://github.com/trytoolchest/toolchest-worker-node/pull/24
- https://github.com/trytoolchest/toolchest-api/pull/22

This does not fix:
- Bigger disk/memory/etc requirements for larger files where args trigger reduced complexity / no parallelization


toolchest_client/tools/tool.py
```diff
@@ -89,70 +90,101 @@ def _prepare_inputs(self):
...
+        ):
+            raise OSError("Output file path must be writable.")
+        if not self.output_name:
+            raise ValueError("Output name must be non-empty.")
+        if self.output_is_directory and not os.path.isdir(self.output_path):
+            raise ValueError(f"Output path must be a directory. It is currently {self.output_path}")
+
+        # Perform a deeper tool_args validation
+        self._validate_tool_args()
+
```
lebovic, line 187: This is the same code block, just moved for clarity

toolchest_client/tools/tool_args.py
```diff
@@ -0,0 +1,326 @@
...
+not parallelization and less validation).
+
+Note: Some tags (e.g., for inputs: "-1", "-2") are filtered out, but
+their functionalities are provided via other arguments (e.g.,
+user-provided inputs).
+"""
+
+VARIABLE_ARGS = "variable"
+
+TOOL_ARG_LISTS = {
```
lebovic, line 28: These are mostly the same args as before, just structured as an explicit whitelist / dangerlist / blacklist.

toolchest_client/tools/tool.py
```diff
@@ -89,70 +90,112 @@ def _prepare_inputs(self):
...
+            raise ValueError("Unknown or blacklisted arguments present in tool_args. See above for details.")
+
+        if dangerous_args:
+            print("WARNING: dangerous arguments found in tool_args. This disables validation and parallelization!")
+            print(f"Dangerous arguments: {dangerous_args}")
+            # Disable parallelization, validation, and revert to plain compressed output
+            self.output_validation_enabled = False
+            self.parallel_enabled = False
+            self.output_is_directory = True
+            self.output_type = OutputType.GZ_TAR
```
bcai2, line 170: Should `self.output_type` have a default value (e.g., `OutputType.FLAT_TEXT`) in `Tool.__init__()`, for backwards compatibility? 

Unless `self.output_type` is explicitly defined for a tool's `SpecificTool.__init__()` function (which it probably should for every tool, just for clarity), it doesn't appear to be assigned a value explicitly unless `dangerous_args` are detected. This could pose a problem for any unpacking logic later.

bcai2 (commented):
Looks good, a few comments.

bcai2 (approved):

tests/test_star.py
```diff
@@ -0,0 +1,61 @@
...
+    )
+
+    toolchest.STAR(
+        read_one=input_file_path,
+        output_path=output_file_path,
+        database_name="GRCh38",
+    )
+
+    # Because STAR is non-deterministic, verify that the number of bytes is the exact same
+    assert os.path.getsize(output_file_path) == 185952766
```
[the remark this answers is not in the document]
lebovic: The output size looks like it's fluctuating by ~22 bytes, so this might have to be adjusted.

That said, the author of STAR claims that "with default parameters and running on one thread, STAR is fully deterministic. With multiple threads, the order of the reads in the output is not deterministic [...] the alignments for each read are [still] completely deterministic.", although that was in a previous version.

toolchest_client/tools/tool_args.py
```diff
@@ -0,0 +1,327 @@
+"""
+toolchest_client.tools.arg_whitelist
```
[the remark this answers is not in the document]
lebovic: 👍

toolchest_client/tools/tool.py
lebovic: Oooh yes, good catch

lebovic: Added a default value of `OutputType.FLAT_TEXT`

toolchest_client/tools/tool_args.py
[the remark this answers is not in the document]
lebovic: Fixed

lebovic:
Tests are actually passing at this point, but are showing as failing due to the CircleCI parallelism split

lebovic:
Looks like CircleCI is being very laggy, but unit tests passed on the second-to-last commit and the final commit was simply a comment change. Merging!

PR #58, opened 2021-12-11
Fix boto3 install

lebovic:
Modifies:
- `boto3` install, so that it's a normal dependency rather than a dev dependency


lebovic:
Merging prior to review, as this is a trivial fix that's blocking testing.

PR #60, opened 2021-12-15
Add support for multipart uploads

lebovic:
Adds:
- Support for multipart uploads via boto3
- Shifts download to boto3 as well

Removes:
- Usage of pre-signed URLs and raw file posting via `requests`


bcai2 (approved):
LGTM, pending review of the API PR.

PR #61, opened 2021-12-15
Increase test tool limit to 256 GB

lebovic:
Modifies:
- Test tool per-file limit from 4.5 GB (default) to 256 GB


lebovic:
Merging without review, as this is a small PR in-line with patterns that's about to be merged

PR #63, opened 2021-12-16
Use per-tool limit rather than single part upload limit to check file sizes

lebovic:
Modifies:
- What values is used to limit per-tool upload sizes


bcai2 (approved):
LGTM

PR #64, opened 2021-12-16
Fix file downloading

lebovic:
Modifies:
- Actually download files

### DIFF
--- a/toolchest_client/api/query.py
+++ b/toolchest_client/api/query.py
@@ -335,7 +335,7 @@ class Query():
                 aws_secret_access_key=output_file_keys["secret_access_key"],
                 aws_session_token=output_file_keys["session_token"],
             )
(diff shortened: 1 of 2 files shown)


toolchest_client/api/query.py
```diff
@@ -335,7 +335,7 @@ def _download(self, output_path):
                 aws_secret_access_key=output_file_keys["secret_access_key"],
                 aws_session_token=output_file_keys["session_token"],
             )
-            s3_client.upload_file(
+            s3_client.download_file(
```
lebovic, line 338: !

bcai2 (approved):
Oops, should've caught that. LGTM, don't forget to bump the version # if this requires it

lebovic:
Failed because CircleCI is down

PR #65, opened 2021-12-16
Increase the per-file limit for STAR

lebovic:
Modifies:
- The STAR per-file limit


bcai2 (approved):
:+1:

bcai2:
Do you foresee this being a concern at all?

lebovic:
I.e. are we expecting people to hit this limit?

PR #68, opened 2021-12-20
Standardize output file interface and enable output file compression across the board

lebovic:
Adds:
- Output file compression across the board, except if parallelism is actively being used.
- Explicit `parallelize=True` flag.
- 32 GB of pure EBS disk for shi7 and an increase limit of 5GB per file. This can be trivially increased.

Modifies:
- Output file format for many tools.

Removes:
- cutadapt (no longer in use)
- shi7 parallelization (no longer needed)
- shogun parallelization (no longer needed)

Associated design doc: https://docs.google.com/document/d/1vn26gPgLHvSqDREpLXoFp8LV_5wDhqWRGXkCBuKTknM/edit#


lebovic:
The second to last commit (and many before) had integration tests enabled

toolchest_client/tools/tool.py
```diff
@@ -75,10 +79,7 @@ def _prepare_inputs(self):
             self.input_files = files_in_path(self.inputs)
             self.num_input_files = len(self.input_files)
 
-        # Sanitize valid paths and list all files in self.inputs.
-        self.input_files = files_in_path(self.inputs)
-
-        self.num_input_files = len(self.input_files)
```
lebovic, line 81: @bcai2 Do you know why the block above was reinserted? It has a git annotation of 12/8/2021.

The same logic should be expressed through the `if self.compress_inputs / else` block above, this variant breaks `cellranger_mkfastq`

toolchest_client/tools/unicycler.py
```diff
@@ -23,4 +25,8 @@ def __init__(self, tool_args, output_name, inputs, input_prefix_mapping,
             input_prefix_mapping=input_prefix_mapping,
             min_inputs=1,
             max_inputs=3,
+            parallel_enabled=False,
+            output_type=OutputType.GZ_TAR,
+            output_is_directory=True,
+            output_names=["assembly.fasta"],
```
lebovic, line 31: Some tools – like Unicycler – have a lot more outputs by default. In the case of Unicycler, we have:

```
001_best_spades_graph.gfa 
003_bridges_applied.gfa 
005_polished.gfa 
assembly.gfa
r1.fastq 
unassembled.fasta
002_overlaps_removed.gfa  
004_final_clean.gfa
assembly.fasta
contigs.fasta.gz
r2.fastq 
unicycler.log
```

All of which I'm fine with keeping in the default outputs, but I'm not sure we want `output_names` to express a _complete_ overview of the output files; some output files change dependent on the inputs and args. Maybe we just try to grab primary output + log?

toolchest_client/tools/star.py
```diff
@@ -30,9 +30,11 @@ def __init__(self, tool_args, output_name, inputs, input_prefix_mapping,
             max_inputs=2,
             database_name=database_name,
             database_version=database_version,
-            parallel_enabled=True,
+            parallel_enabled=True if parallelize else False,
```
bcai2, line 33: Thoughts on `parallel_enabled=parallelize`?

toolchest_client/tools/star.py
```diff
@@ -30,9 +30,11 @@ def __init__(self, tool_args, output_name, inputs, input_prefix_mapping,
...
             max_input_bytes_per_file=128 * 1024 * 1024 * 1024,
-            output_type=OutputType.SAM_FILE,  # In many cases, this will change (e.g. a dangerous argument is passed)
+            max_input_bytes_per_file_parallel=4.5 * 1024 * 1024 * 1024,
+            output_type=OutputType.SAM_FILE if parallelize else OutputType.GZ_TAR,
+            output_is_directory=False if parallelize else True,
```
bcai2, line 37: ^, `output_is_directory=not parallelize`?

toolchest_client/tools/tool.py
bcai2: Seems like it was a merge commit gone wrong (my bad!).

bcai2 (approved):
:+1:, see comments

toolchest_client/tools/tool.py
lebovic: 👍

toolchest_client/tools/star.py
lebovic: That works too! In the back of my mind I was thinking of the user input as dirty and potentially non-boolean given that we don't have types implemented.

I have a slight preference for this, but am open to changing if you have a preference.

toolchest_client/tools/star.py
lebovic: Updated

lebovic: (different than the above from my perspective, as this includes an implicit boolean cast)

toolchest_client/tools/star.py
lebovic: Merging in the meantime as this would be a trivial fix and I want to get the integration tests started

PR #69, opened 2021-12-20
S3 chaining

lebovic:
Adds:
- Output class returned by all `toolchest.tool()` calls, which contains `s3_uri`, `presigned_s3_url`, and (local) `output_path` variables
- S3 chaining, via supplying `output.s3_uri` from a previous tool as the `inputs` parameter for a following tool
- the ability to skip download of any tool's output, by setting `output_path=None` (set to `None` by default)


bcai2:
Pending adding integration tests, but manual tests are working.

tests/test_chaining.py
```diff
@@ -0,0 +1,6 @@
+import os
+import pytest
+
+import toolchest_client as toolchest
+
+THIS_DIRECTORY = os.path.dirname(os.path.realpath(__file__))
```
lebovic, line 6: Is this here as a placeholder for integration tests? Are you planning to add before you merge?

toolchest_client/api/output.py
```diff
@@ -0,0 +1,31 @@
...
+    def __init__(self, s3_uri=None, presigned_s3_url=None, output_path=None):
+        self.s3_uri = s3_uri
+        self.presigned_s3_url = presigned_s3_url
+        self.output_path = output_path
+
+    def __repr__(self):
+        return str(self.__dict__)
+
+    def __str__(self):
+        return str(self.__dict__)
```
lebovic, line 31: Nit: missing newline at EOF

toolchest_client/api/query.py
```diff
@@ -50,6 +51,10 @@ def __init__(self):
         self.thread_name = ''
         self.thread_statuses = None
 
+        self.presigned_s3_url = None
+        self.unpacked_output = None
+        self.output = output_object if output_object else Output()
```
lebovic, line 56: Is there a case in which the `output_object` is passed during initialization?

lebovic: (and where we want to set `self.output` with it?)

toolchest_client/tools/tool.py
```diff
@@ -180,16 +182,14 @@ def _validate_args(self):
...
-        if not os.access(
+        if self.output_path and not os.access(
                 os.path.dirname(self.output_path),
                 os.W_OK | os.X_OK,
         ):
             raise OSError("Output file path must be writable.")
         if not self.output_name:
             raise ValueError("Output name must be non-empty.")
-        if self.output_is_directory and not os.path.isdir(self.output_path):
+        if self.output_is_directory and self.output_path and not os.path.isdir(self.output_path):
```
lebovic, line 192: Out of curiosity, why was this rewritten?

toolchest_client/tools/tool.py
```diff
@@ -208,25 +208,31 @@ def _preflight(self):
...
-                    raise ValueError(f"Output file at {output_file_path} is suspiciously small")
+        if self.output_path:
+            if self.output_validation_enabled:
+                for output_name in self.output_names:
+                    output_file_path = f"{self.output_path}/{output_name}"
+                    assert_exists(output_file_path, must_be_file=True)
+                    if os.stat(output_file_path).st_size <= 5:
+                        raise ValueError(f"Output file at {output_file_path} is suspiciously small")
+        else:
+            print("Output path not provided, skipping download.")
```
lebovic, line 235: 👍

toolchest_client/tools/tool.py
```diff
@@ -381,14 +387,17 @@ def run(self):
...
         for index, input_files in enumerate(jobs):
             # Add split files for merging and later deletion, if running in parallel
             temp_parallel_output_file_path = f"{self.output_path}_{index}"
             if should_run_in_parallel:
                 temp_input_file_paths += input_files
                 temp_output_file_paths.append(temp_parallel_output_file_path)
-            q = Query()
+
+            # Create a new Output for the thread.
+            self.output_locations[index] = Output()
```
lebovic, line 399: Nit: `output_object` and `output_locations` are hard to parse for someone – maybe even us – reading this code 6 months from now.

Is there a way you could rename this to be more specific (e.g. to provide meaning to what the `[index]` is, that the `output_location`s are tied to threads, etc)?

lebovic (commented):
A couple nits, otherwise looks good!

toolchest_client/tools/tool.py
bcai2: This was rewritten to allow for `output_path=None`, which skips the local download.

tests/test_chaining.py
bcai2: Precisely; I'll add tests in a follow-up commit to this PR.

toolchest_client/api/query.py
bcai2: Yes; when multithreading (which all of our jobs are currently doing), an `Output` is initialized before the `Thread()` call for each thread and stored in the `Tool.thread_outputs` dict. This is to provide persistence after `Query.run_query()` executes.

The behavior is currently to return the `Output` from the first thread, which might not make sense because its `output_path` leads to a temporary file outputted from a single thread. I'll mark this as an issue for the future, since we aren't currently using multithreading and may change threading behavior in the future fundamentally.

toolchest_client/tools/tool.py
bcai2: Done, hopefully the refactored names are more clear now.

(Note: `Output.output_path` may be a bit unclear; this refers to the locally downloaded output path(s), if there are any. I didn't specify it as `local_output_path` because I think `output_path` on its own is straightforward enough for the user, but we can amend that if needed.)

bcai2:
Pre-emptively merging this in to let integration tests run + progress on the user-specified download interface, which branches from here.

PR #70, opened 2021-12-20
[Hotfix] Shift burden of unpack structure to the worker node

lebovic:
Modifies:
- Removes the expectation that there will be a `output/` component of `output.tar.gz`


bcai2 (approved):
LGTM; I'll have to merge this to staging and then to my dev branch since this is exactly what I'm working on (in passing the unpacked filenames to the Output object).

PR #72, opened 2021-12-28
Propagate order of inputs within tool arg prefix

bcai2:
Adds:
- Propagation of the order of each input within each tool prefix for `megahit`. This order is 0-indexed and will be passed to the API when uploading/registering each input file. 
 
For example, `read_one = ["input_one", "input_two", "input_three"]` will result in: 
```
  input_prefix_mapping["input_one"]["order"]: 0
  input_prefix_mapping["input_two"]["order"]: 1
  input_prefix_mapping["input_three"]["order"]: 2
```

Note: the order value is stored as the `"order"` parameter within `tool_prefix_mapping`., though it is eventually passed as `tool_prefix_order` to the API.

Requires:
- https://github.com/trytoolchest/toolchest-api/pull/31
- https://github.com/trytoolchest/toolchest-worker-node/pull/30

### DIFF
--- a/toolchest_client/tools/tool.py
+++ b/toolchest_client/tools/tool.py
@@ -30,8 +30,8 @@ class Tool:
     def __init__(self, tool_name, tool_version, tool_args, output_name,
                  output_path, inputs, min_inputs, max_inputs,
                  database_name=None, database_version=None,
-                 input_prefix_mapping=None, parallel_enabled=False,
-                 max_input_bytes_per_file=FOUR_POINT_FIVE_GIGABYTES,
+                 input_prefix_mapping=None, input_prefix_order=None,
+                 parallel_enabled=False, max_input_bytes_per_file=FOUR_POINT_FIVE_GIGABYTES,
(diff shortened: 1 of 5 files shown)


toolchest_client/tools/tool.py
```diff
@@ -400,6 +401,7 @@ def run(self):
                 "output_name": f"{index}_{self.output_name}" if should_run_in_parallel else self.output_name,
                 "input_files": input_files,
                 "input_prefix_mapping": self.input_prefix_mapping,
+                "input_prefix_order": self.input_prefix_order,
```
lebovic, line 404: I think it makes sense to add the input prefix to `input_prefix_mapping`. For an example, see the worker node PR

lebovic (changes requested):

toolchest_client/tools/tool.py
bcai2: Gotcha, I'll do a quick refactor

lebovic (approved):

PR #74, opened 2021-12-30
Remove limit on # of input files

bcai2:
Adds:
- Capability of tools to have an unlimited number of input files, by specifying `max_inputs=None` in the class `__init__()` function. 

(Note: `max_inputs` is set to `None` by default if left unspecified during `__init__()`, but all of our current tools specify `max_inputs` explicitly.)

Removes:
- The arbitrary cap of 10 input files for `megahit`.
- The arbitrary cap of 10 input files for `shi7`. 

(Both tools can now have an unlimited number of inputs.)

### DIFF
--- a/toolchest_client/tools/tool.py
+++ b/toolchest_client/tools/tool.py
@@ -28,7 +28,7 @@ FOUR_POINT_FIVE_GIGABYTES = int(4.5 * 1024 * 1024 * 1024)
 
 class Tool:
     def __init__(self, tool_name, tool_version, tool_args, output_name,
-                 output_path, inputs, min_inputs, max_inputs,
+                 output_path, inputs, min_inputs, max_inputs=None,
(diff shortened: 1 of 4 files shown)


toolchest_client/tools/tool.py
```diff
@@ -85,7 +85,8 @@ def _prepare_inputs(self):
         if self.num_input_files < self.min_inputs:
             raise ValueError(f"Not enough input files submitted. "
                              f"Minimum is {self.min_inputs}, {self.num_input_files} found.")
-        if self.num_input_files > self.max_inputs:
+        # If max_inputs = None, this check is ignored.
```
lebovic, line 88: Nit: I don't think this comment is necessary, it's pretty clear by the `if self.max_inputs and`

lebovic (approved):

PR #76, opened 2022-01-05
Refactor downloads, add download interface

bcai2:
Adds:
* `toolchest.download(output_path, s3_uri="s3://your-s3-uri")` and `output.download(output_dir)`, user-invoked functions which can download an output after its job has completed.
* `ToolchestDownloadException`, a new exception thrown whenever an error is encountered in downloading or unpacking.
* Integration tests for manual downloads.

Modifies:
* Code for downloading has now been moved from `Query` to `api/downloads.py`, which serves as a single source for all download functionality.
* Downloads during `Query.run_query()` now call the functions in `api/downloads.py`.
* URLs used for the API have been moved to `api/urls.py`, which serves as a single source for all API URLs.
* All `toolchest.tool()` calls now correctly return an `Output`.

Removes:
* `presigned_s3_url` parameter for `Output`

### DIFF
--- a/tests/test_download.py
+++ b/tests/test_download.py
@@ -17,7 +17,7 @@ def test_kraken2_output_manual_download():
     Tests Kraken 2 against the standard (v1) database, with
     output manually downloaded after job completion
     """
-    test_dir = "test_kraken2_standard"
+    test_dir = "test_kraken2_output_manual_download"
     os.makedirs(f"./{test_dir}", exist_ok=True)
     input_file_s3_uri = "s3://toolchest-integration-tests-public/synthetic_bacteroides_reads.fasta"
     output_dir_path = f"./{test_dir}/"
--- a/tests/test_download.py
+++ b/tests/test_download.py
@@ -22,8 +22,10 @@ def test_kraken2_output_manual_download():
     input_file_s3_uri = "s3://toolchest-integration-tests-public/synthetic_bacteroides_reads.fasta"
     manual_output_dir_path = f"./{test_dir}/manual/"
     manual_output_file_path = f"{manual_output_dir_path}kraken2_output.txt"
-    toolchest_output_dir_path = f"./{test_dir}/toolchest/"
-    toolchest_output_file_path = f"{toolchest_output_dir_path}kraken2_output.txt"
+    toolchest_s3_dir_path = f"./{test_dir}/toolchest/s3/"
+    toolchest_s3_file_path = f"{toolchest_s3_dir_path}kraken2_output.txt"
+    toolchest_pipeline_dir_path = f"./{test_dir}/toolchest/id/"
+    toolchest_pipeline_file_path = f"{toolchest_s3_dir_path}kraken2_output.txt"
 
     # Run job without downloading
     output = toolchest.kraken2(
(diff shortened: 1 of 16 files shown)


bcai2:
@lebovic Ready to review, whenever you're at a good stopping point. Otherwise, I'll pre-emptively merge this into staging and respond to changes in a new PR.

tests/test_download.py
```diff
@@ -0,0 +1,68 @@
...
+
+toolchest_api_key = os.environ.get("TOOLCHEST_API_KEY")
+if toolchest_api_key:
+    toolchest.set_key(toolchest_api_key)
+
+KRAKEN2_SINGLE_END_HASH = 886254946
+
+
+@pytest.mark.integration
+def test_kraken2_output_manual_download():
```
lebovic, line 15: About how long does this test take?

toolchest_client/api/download.py
```diff
@@ -0,0 +1,127 @@
...
+
+from toolchest_client.api.auth import get_headers
+from toolchest_client.api.exceptions import ToolchestDownloadError
+from toolchest_client.api.urls import PIPELINE_URL
+from toolchest_client.files import get_file_type, get_params_from_s3_uri, unpack_files
+
+
+def download(output_path, s3_uri=None, pipeline_segment_instance_id=None,
+             output_file_keys=None, output_type=None, ):
+    """Downloads output to ``output_path``.
```
lebovic, line 27: Possible nit: unintentional "``"?

toolchest_client/api/download.py
```diff
@@ -0,0 +1,127 @@
...
+            # Note: this assumes the pipeline segment instance ID is embedded in the egress S3 URI.
+            # This supersedes pipeline_segment_instance_id, if it is provided.
+            s3_uri_params = get_params_from_s3_uri(s3_uri)
+            pipeline_segment_instance_id = s3_uri_params["key_initial"]
+        if pipeline_segment_instance_id:
+            output_s3_uri, output_file_keys = get_download_details(pipeline_segment_instance_id)
+        else:
+            error_message = "S3 URI of output not provided."
+            raise ToolchestDownloadError(error_message) from None
+    print(output_file_keys)
```
lebovic, line 58: Nit: leftover `print`?

toolchest_client/files/tests/test_s3.py
```diff
@@ -5,12 +5,14 @@
 
 
 def test_s3_params():
-    example_s3_uri = "s3://toolchest-public-examples/example.fastq"
+    example_s3_uri = "s3://toolchest-public-examples/dummy-id/example.fastq"
     params = get_params_from_s3_uri(example_s3_uri)
     target_params = {
-        "arn": "arn:aws:s3:::toolchest-public-examples/example.fastq",
+        "arn": "arn:aws:s3:::toolchest-public-examples/dummy-id/example.fastq",
```
lebovic, line 11: Worth noting that we're now expecting the uuid to be present in the S3 URI structure. This isn't reflected in the db schema rn, where the uuid is just a part of the file prefix.

lebovic (approved):
👍

toolchest_client/api/download.py
bcai2: That was a holdover from the previous comment in `query.py`, but it does seem unnecessary, so I'll get rid of it.

toolchest_client/api/download.py
bcai2: Yup, good catch

tests/test_download.py
bcai2: Not too much longer than the `kraken2` single-end run (~3 mins). This is basically just the same as the normal `kraken2` single-end run, but downloading the output in 3 different ways after the job finishes.

PR #77, opened 2022-01-06
v0.7.51

bcai2:
Adds:
- propagation of order inputs within a given tool arg prefix (#72)
- `toolchest.tool()` now returns an `Output` that contains information about output file locations (#69)
- support for tool chaining (using the output of one job as the input for another),  via the `Output` returned from a `toolchest.tool()` call (#69)
- `toolchest.download()`, an interface for manually downloading outputs of completed jobs
- integration tests for `megahit` (#78)

Modifies:
- limit on # of tool input files, which is no longer necessary (#74)
  - limit removed for `shi7` and `megahit`
- S3 file validation, which is now server-side instead of client-side (#79)

Requires:
- https://github.com/trytoolchest/toolchest-worker-node/pull/34
- https://github.com/trytoolchest/toolchest-api/pull/35


bcai2:
^ failing due to lack of version bump on staging. no significant changes, merging to prod now

PR #78, opened 2022-01-08
Megahit integration tests + integration test bugfix

bcai2:
Adds:
- Two integration tests for `megahit`:
  - One uses two interleaved inputs, one pair of paired-end inputs, and two single-end inputs. This matches the integration test used in the actual `megahit` package.
  - The other uses two different-sized pairs of paired-end inputs, to test that the API properly processes pairs.

Both integration tests run in 2-4 minutes.

Modifies:
- Bugfix for failing integration tests not using the manual download.
- Bugfix for `megahit` inputs that are given as strings instead of lists of strings.

### DIFF
--- a/tests/test_megahit.py
+++ b/tests/test_megahit.py
@@ -8,21 +8,25 @@ toolchest_api_key = os.environ.get("TOOLCHEST_API_KEY")
 if toolchest_api_key:
     toolchest.set_key(toolchest_api_key)
 
+EXPECTED_MIN_OUTPUT_SIZE = 1000
+
+
 @pytest.mark.integration
-def test_megahit_one_pair():
+def test_megahit_many_types():
     """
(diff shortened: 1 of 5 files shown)


tests/test_megahit.py
```diff
@@ -0,0 +1,74 @@
...
+
+@pytest.mark.integration
+def test_megahit_many_types():
+    """
+    Tests Megahit with two interleaved inputs, one pair of paired-end inputs,
+    and two single-end inputs.
+
+    Note: Multithreaded megahit is not deterministic, so
+    we check the size of the file instead.
+    See https://github.com/voutcn/megahit/issues/48.
```
lebovic, line 22: Predictable, but also oof!

lebovic (approved):
LGTM 👍

PR #34, opened 2021-08-27
Add a few more integration tests

lebovic:
Adds automated integration tests for:
- Unicycler
- STAR
- Parallel STAR

Still needs:
- Resolve CircleCI 10 min timeout

PR #79, opened 2022-01-12
Refactor S3 file size check

bcai2:
Adds:
- `get_s3_file_size(uri)` in `files/s3.py`, allowing the client to check the size of an accessible file in S3 via the API.
- S3 base routes to `api/urls.py`.

Modifies:
- `assert_accessible_s3`, which is refactored to be a validation of `get_s3_file_size()`.
- Size checking of S3 files is now based on querying the API via `get_s3_file_size(uri)`, instead of trying to check directly via `boto3`.

Requires:
- https://github.com/trytoolchest/toolchest-api/pull/34


lebovic (approved):
👍  LGTM

PR #82, opened 2022-01-25
fix: update pysam to latest version

jherr-dev:
Updated pysam so poetry install works. I do have cython installed as well so that may still be a requirement in addition to the version update.


bcai2 (approved):
:+1: (pending a test on a fresh install?)

lebovic:
I'm having trouble effectively testing a clean install, but LGTM otherwise. Let's gate on a fresh install, I think Bryce has a fresh MacBook to try on?

Longer term, we should add better tests for different environment installs.

lebovic:
Also – can you re-enable the parallel STAR test?

PR #85, opened 2022-01-28
feat: add cellranger_count

bcai2:
Adds:
- User interface for `cellranger_count`. The `inputs` argument must be a directory containing the input FASTQs. The directory is then zipped and uploaded.
- Support for S3 inputs for `cellranger_count`. S3 inputs must be zipped archives.
- Integration tests for `cellranger_count`, testing with inputs 
- Internal option to skip decompression in `toolchest.download()`.
- A second (MD5) hash function, used for integration tests on binary files.
- Custom Windows error message during unpacking outputs if long path lengths are not enabled.

Modifies:
- Checking whether a path is an S3 URI is now single-sourced with `path_is_s3_uri()` in `files/s3.py`.

Removes:
- `cellranger_mkfastq` interface.

Note: The minor version number has been bumped to `0.8.0`.


tests/test_cellranger.py
```diff
@@ -0,0 +1,77 @@
...
+    )
+    shutil.unpack_archive(packed_inputs_path, input_dir_path)
+    output = toolchest.cellranger_count(
+        inputs=input_dir_path,
+        transcriptome_name="GRCh38",
+    )
+    verify_cellranger_count_outputs(output, output_dir_path)
+
+
+def verify_cellranger_count_outputs(output, output_dir_path):
```
lebovic, line 53: Is there a benefit to this being a separate function from `test_cellranger_count`? Looks like this isn't generalizable (e.g. locked hashes)

tests/test_cellranger.py
```diff
@@ -0,0 +1,77 @@
...
+if toolchest_api_key:
+    toolchest.set_key(toolchest_api_key)
+
+MIN_EXPECTED_ARCHIVE_SIZE = 34000000
+MAX_EXPECTED_ARCHIVE_SIZE = 38000000
+
+EXPECTED_SUMMARY_SIZE = 2744825
+EXPECTED_RAW_MATRIX_SIZE = 868393
+EXPECTED_RAW_MATRIX_HASH = "d00cca1d2b4344b03946eeaeedc17ed5"
+EXPECTED_FILTERED_MATRIX_SIZE = 503956
```
lebovic, line 18: Is there a benefit to abstracting these as global constants rather than just having them locally in the assertion?

tests/util/hash.py
```diff
@@ -8,9 +9,20 @@ def unordered(file_path):
     """
     file_hash = 1
     eighth_mersenne_prime = 2147483647
-    with open(file_path) as file:
+    with open(file_path, errors="ignore") as file:
```
lebovic, line 12: Out of curiosity, why `errors="ignore"`?

tests/util/hash.py
```diff
@@ -8,9 +9,20 @@ def unordered(file_path):
...
+
+
+def binary_hash(file_path):
+    """
+    Generates an MD5 hash of a binary file.
+    """
+    with open(file_path, "rb") as file:
+        print("Hashing", file_path)
+        file_hash = hashlib.md5(file.read()).hexdigest()
+    print("Hash is", file_hash)
```
lebovic, line 27: Fyi, the only reason the `print` statements were included in the `adler32` hash above is because I ported this over from a Jupyter notebook and missed that cleanup.

Not sure if print statements add any benefit here ¯\_(ツ)_/¯

toolchest_client/api/download.py
```diff
@@ -121,5 +125,9 @@ def _unpack_output(compressed_output_path, output_type=None):
         )
     except Exception as err:
         error_message = f"Failed to unpack file with type: {output_type}."
+        if sys.platform == "win32":
+            PATH_TOO_LONG_CODE = 206
+            if isinstance(err, FileNotFoundError) and err.winerror == PATH_TOO_LONG_CODE:
+                error_message += "\nLong file name support in Windows 10 must be enabled."
```
lebovic, line 131: Wild

toolchest_client/files/general.py
```diff
@@ -38,8 +38,7 @@ def check_file_size(file_path, max_size_bytes=None):
     :param max_size_bytes: Maximum number of bytes allowed for a file. Throws error if above limit.
     :type max_size_bytes: int | None
     """
-    S3_PREFIX = "s3://"
-    if not file_path.startswith(S3_PREFIX):
+    if not path_is_s3_uri(file_path):
```
lebovic, line 41: 👍

toolchest_client/tools/api.py
```diff
@@ -44,35 +43,36 @@ def bowtie2(inputs, output_path=None, database_name="GRCh38_noalt_as", database_
...
 
-def cellranger_mkfastq(inputs, samplesheet_name, output_path=None, tool_args=""):
-    """Runs Cell Ranger's mkfastq command via Toolchest.
+def cellranger_count(inputs, transcriptome_name="GRCh38", output_path=None, tool_args=""):
+    """Runs Cell Ranger's count command via Toolchest.
 
     :param inputs: Path (client-side) to be passed in as input.
     :param output_path: (optional) Path (client-side) where the output file will be downloaded.
-    :param samplesheet_name: Name of sample sheet. Expected to exist inside of "inputs".
+    :param transcriptome_name: Name of transcriptome. Defaults to `GRCh38`.
```
lebovic, line 51: Huh, interesting. Technically the database for Bowtie and some others would be a `genome_name`. Up to you how you name it – I see the benefit to keeping this consistent with the Cell Ranger interface, but I also see a risk that we start making other params for other tools very specific when they're not needed.

toolchest_client/tools/tool.py
```diff
@@ -76,8 +76,12 @@ def __init__(self, tool_name, tool_version, tool_args, output_name,
...
         """Prepares the input files."""
         if self.compress_inputs:
-            # Input files are all .tar.gz'd together, preserving directory structure
-            self.input_files = [compress_files_in_path(self.inputs)]
+            if path_is_s3_uri(self.inputs):
+                # If the given path is in S3, it is assumed to be compressed already.
+                self.input_files = [self.inputs]
+            else:
+                # Input files are all .tar.gz'd together, preserving directory structure
+                self.input_files = [compress_files_in_path(self.inputs)]
```
lebovic, line 84: Huh, cool

lebovic (approved):
LGTM, some small comments

tests/test_cellranger.py
bcai2: This is separated into its own function to prevent copy-pasting the same code. For scoping, we can bring this within `test_cellranger_count()`.

tests/test_cellranger.py
bcai2: This is the format that we've been following for the other tests, but we can move them locally into `verify_cellranger_count_outputs()`.

tests/util/hash.py
bcai2: They generally help for manually invoking the hashes during testing (which is the only place hashes are being used), so I'll keep them in for now.

tests/util/hash.py
bcai2: This is to prevent unicode errors from happening when hashing non-ASCII encoded files. However, it isn't being actively used (since binary files are being hashed with MD5), so I can remove it.

toolchest_client/tools/api.py
bcai2: I would keep this consistent with the cellranger interface for now (with the limited number of tools we have), but I'm open to changing this down the road.

tests/test_cellranger.py
lebovic: Ah I missed that it was used twice.

Adding followup comment, then.

tests/test_cellranger.py
lebovic: I started this pattern with Kraken 2 because it needed to be referenced twice, outside the scope of the same function (https://github.com/trytoolchest/toolchest-client-python/blob/main/tests/test_kraken2.py).

It's harder to reason about these things at a glance if they're abstracted / moved to the non-relevant part of the file.

lebovic: If it exists improperly in other tests, that's my fault and and I should have caught it earlier.

tests/util/hash.py
lebovic: Defs remove if unnecessary

tests/util/hash.py
lebovic: 👍

toolchest_client/tools/api.py
lebovic: I'd argue that keeping this consistent with the cellranger interface is analogous to maintaining capitalization to be consistent with the naming of the tools on the command line – it keeps consistency for those who are moving CLI -> Toolchest, but it decreases consistency within Toolchest.

I think it's important to think about who is actually going to be using this, and why. We don't have good data yet on who will be coming from CLI vs. using the tool for the first time in Toolchest / using multiple tools on Toolchest, so I can't really make a case – but worth keeping in mind that this comes across as an inconsistency within Toolchest.

lebovic: (I'm okay with not changing this)

tests/test_cellranger.py
bcai2: Sounds good, moving to `verify_cellranger_count_outputs()`

toolchest_client/tools/api.py
bcai2: I'll change it to `database_name` to be in line with our other tools, then.

tests/test_cellranger.py
```diff
@@ -0,0 +1,77 @@
...
+
+    # Test using a compressed archive in S3
+    output = toolchest.cellranger_count(
+        inputs="s3://toolchest-integration-tests/cellranger/count/pbmc_1k_v3_fastqs_trimmed.tar.gz",
+        transcriptome_name="GRCh38",
+    )
+    verify_cellranger_count_outputs(output, output_dir_path)
+
+    shutil.rmtree(output_dir_path)
+    os.makedirs(output_dir_path, exist_ok=True)
```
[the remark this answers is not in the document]
bcai2: :+1:, refactored

PR #88, opened 2022-01-31
chore: format to flake8

jherr-dev:


.flake8
```diff
@@ -0,0 +1,8 @@
+[flake8]
+max-line-length = 120
+exclude=
+    demo.py
+    docs/conf.py
+per-file-ignores =
+    # ignores unused imports and imports not at the top of file in init files
+    */__init__.py:F401,E402
```
bcai2, line 8: If adding a newline doesn't break flake8, then I would add one here. Otherwise, looks good.

PR #90, opened 2022-02-02
feat: update shi7 integration tests

jherr-dev:
Marks shi7 tests as ready to run and updates output directory variables.


bcai2 (approved):
:+1:

PR #91, opened 2022-02-02
fix: change shi7 test directories

jherr-dev:
Simplifies the test directories for shi7 integration tests.


bcai2 (approved):
:+1:, though my recommendation would be to separate the input and output directories, but if tests are passing, then I'm good with this

bcai2 (approved):
:+1:

PR #95, opened 2022-02-05
chore: refactor `PIPELINE_URL` and `_add_input_file()`

bcai2:
Modifies:
*  In `api/query.py`, renames `_add_input_file()` to `_register_input_file()`.
* Deprecates including `is_s3` in `_register_input_file()`, which is unnecessary with `s3_uri`.
* Renames the confusing variable `PIPELINE_URL` to `PIPELINE_SEGMENT_INSTANCES_URL`.
* Removes stray comments in `_register_input_file()`.

Note: This can be merged into #94 directly. We should let integration tests run if merging to staging instead, but I'd hold off on updating to `0.8.1` until #94 is merged.

### DIFF
--- a/toolchest_client/api/download.py
+++ b/toolchest_client/api/download.py
@@ -19,7 +19,7 @@ from requests.exceptions import HTTPError
 
 from toolchest_client.api.auth import get_headers
 from toolchest_client.api.exceptions import ToolchestDownloadError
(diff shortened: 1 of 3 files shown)


jherr-dev (approved):
I would change base to #94 since this can wait to be merged until that is but otherwise lgtm.

PR #96, opened 2022-02-09
feat: add support for outputting to an s3 bucket

jherr-dev:
Allows the output path of a tool to be an s3 bucket that the worker nodes have permission to access.  Also adds s3 output type to kraken2 which will skip output compression.

### DIFF
--- a/toolchest_client/api/query.py
+++ b/toolchest_client/api/query.py
@@ -85,6 +85,7 @@ class Query:
             tool_name=tool_name,
             tool_version=tool_version,
             tool_args=tool_args,
+            output_file_path=output_path
         )
         create_content = create_response.json()
 
@@ -121,7 +122,7 @@ class Query:
 
     def _send_initial_request(self, tool_name, tool_version, tool_args,
                               database_name, database_version, output_name,
-                              compress_output):
+                              compress_output, output_file_path):
         """Sends the initial request to the Toolchest API to create the query.
 
         Returns the response from the POST request.
@@ -135,6 +136,7 @@ class Query:
             "output_file_name": output_name,
             "tool_name": tool_name,
             "tool_version": tool_version,
+            "output_file_path": output_file_path,
         }
 
         create_response = requests.post(
@@ -339,7 +341,7 @@ class Query:
 
         try:
             self.output_s3_uri, output_file_keys = get_download_details(self.PIPELINE_SEGMENT_INSTANCE_ID)
-            if output_path:
+            if output_path and not path_is_s3_uri(output_path):
                 self._update_thread_status(ThreadStatus.DOWNLOADING)
                 self._update_status(Status.TRANSFERRING_TO_CLIENT)
                 self.unpacked_output_paths = download(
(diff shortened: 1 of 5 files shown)


toolchest_client/tools/tool.py
```diff
@@ -189,14 +189,15 @@ def _validate_args(self):
 
         if self.inputs is None:
             raise ValueError("No input provided.")
-        if self.output_path and not os.access(
+        if self.output_path and not path_is_s3_uri(self.output_path) and not os.access(
```
lebovic, line 192: I've seen `if self.output_path and not path_is_s3_uri(self.output_path)` five times in this PR – might be worth specifying an instance attribute of `has_local_output` or similar?

lebovic (approved):
One nit, otherwise 👍

toolchest_client/tools/tool.py
jherr-dev: We've already removed `is_s3` as an input file attribute which seems very similar to `has_local_output` I could extract it to a function so it's only fully expanded in 1 spot. Also allows for us to change that condition compared to an attribute that's more locked in.

jherr-dev: We've already removed `is_s3` as an attribute of input file which seems very similar to `has_local_output`. I can extract it to a function instead of an attribute so it's only fully expanded in one spot instead. A function is also more flexible if we want to update the condition in the future.

lebovic: Sounds good – I trust your judgement on either, I also think it would be okay to leave as-is

toolchest_client/api/query.py
```diff
@@ -339,7 +341,7 @@ def _download(self, output_path, output_type):
 
         try:
             self.output_s3_uri, output_file_keys = get_download_details(self.PIPELINE_SEGMENT_INSTANCE_ID)
-            if output_path:
+            if output_path and not path_is_s3_uri(output_path):
```
bcai2, line 344: Adding this condition seems like it should skip the need to edit `unpack_files()` in `unpack.py`; see the other comment.

toolchest_client/files/unpack.py
```diff
@@ -19,6 +20,8 @@ def unpack_files(file_path_to_unpack, output_type):
         return file_path_to_unpack
     elif output_type == OutputType.SAM_FILE:
         return file_path_to_unpack
+    elif output_type == OutputType.S3:
+        return file_path_to_unpack
```
bcai2, line 24: Nit: Setting this definition here seems confusing since the purpose of `unpack_files()` is to unpack compressed outputs. `unpack_files()` is only used in the context of `_unpack_output()`, a helper function for `download()` in `api/download.py`, and `download()` gets skipped if the output path is an S3 URI. I would keep the new `OutputType` but remove these two lines for clarity. (Let me know if this breaks tests, though.)

bcai2 (commented):
Looks good in general, see comments

toolchest_client/files/unpack.py
jherr-dev: This is primarily here for completeness and could be removed (I'm thinking in switch terms and this if chain could be switched to one). However, I do see a possibility where the download prevention is removed if someone wanted to say move a file from our egress bucket to a bucket in their account after a run is finished if they didn't supply an output path at the tool call.

toolchest_client/api/query.py
jherr-dev: This is mainly to prevent the need to update `download()` and may change in the future per my other comment.

PR #98, opened 2022-02-10
feat: create folders along given output path

bcai2:
Adds:
- Output path directory creation if `output_path` does not exist, both in tool runs and manual download.

Modifies:
- `_validate_args` in the `Tool` class no longer checks existence of the output directory.
- Integration tests (where applicable) no longer create an output dir before running a tool or downloading.
- `_is_local_path` is renamed to `_output_path_is_local`

### DIFF
--- a/toolchest_client/api/download.py
+++ b/toolchest_client/api/download.py
@@ -33,8 +33,7 @@ def download(output_path, s3_uri=None, pipeline_segment_instance_id=None,
     `get_download_details()`.
 
     :param output_path: Output path to which the file(s) will be downloaded.
-        This should be a directory that already exists, but direct filenames
-        are also supported.
+        This should be a directory, but direct filenames are also supported.
     :param s3_uri: URI of file contained in S3. This can be passed from
         the parameter `output.s3_uri` from the `output` returned by a previous
         job.
@@ -45,6 +44,7 @@ def download(output_path, s3_uri=None, pipeline_segment_instance_id=None,
     :param skip_decompression: Whether to skip decompression of the downloaded file archive.
     :param output_type: Output type of the produced output file. Used internally.
     """
+    # TODO: support output_path directory creation while supporting non-directory filetypes
 
     if output_file_keys is None:
         if s3_uri:
--- a/toolchest_client/api/download.py
+++ b/toolchest_client/api/download.py
@@ -33,7 +33,7 @@ def download(output_path, s3_uri=None, pipeline_segment_instance_id=None,
     `get_download_details()`.
 
     :param output_path: Output path to which the file(s) will be downloaded.
-        This should be a directory, but direct filenames are also supported.
+        This should be a local directory, but direct filenames are also supported.
     :param s3_uri: URI of file contained in S3. This can be passed from
         the parameter `output.s3_uri` from the `output` returned by a previous
         job.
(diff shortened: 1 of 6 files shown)


toolchest_client/api/download.py
```diff
@@ -58,6 +57,13 @@ def download(output_path, s3_uri=None, pipeline_segment_instance_id=None,
             error_message = "S3 URI of output not provided."
             raise ToolchestDownloadError(error_message) from None
 
+    # Create output directories if output_path does not exist.
+    # Note: Assumes that output_path is a directory if it does not exist.
+    # This may lead to undesired leaf dir creation if output_path is not intended to be a dir,
```
lebovic, line 62: What do you think about a warning / error if the leaf of the path has an extension?

I know that would block some valid paths (e.g. test/actually.a.dir), but would probably make usage more clear

toolchest_client/api/download.py
```diff
@@ -58,6 +57,13 @@ def download(output_path, s3_uri=None, pipeline_segment_instance_id=None,
...
+    # but support for non-dir output_path will likely be deprecated.
```
lebovic, line 63: I'm okay with this tradeoff 👍

lebovic (approved):

toolchest_client/api/download.py
bcai2: I'll add a warning message. (This will only be invoked by manual downloads since directory creation in tool runs is handled before downloading, so the warning message won't get in the way of any tool run output.)

toolchest_client/tools/tool.py
```diff
@@ -222,21 +219,23 @@ def _preflight(self):
...
                 if os.path.exists(self.output_path):
                     if os.path.isfile(self.output_path):
                         raise ValueError(
                             f"{self.output_path} is a file. Please pass a directory instead of an output file."
                         )
                 else:
-                    os.makedirs(self.output_path)
+                    os.makedirs(self.output_path, exist_ok=True)
+            else:
+                os.makedirs(os.path.dirname(self.output_path), exist_ok=True)
```
jherr-dev, line 232: Why are you making a dir if the output is not a directory here?

bcai2: Based on `os.path.dirname`, this command makes directories up until the file name (e.g., if the path is `/path/to/output/file.txt`, then `/path/to/output` gets created as a directory).

jherr-dev: Makes sense just skimmed a bit too much.

toolchest_client/tools/tool.py
```diff
@@ -35,13 +35,13 @@ def __init__(self, tool_name, tool_version, tool_args, output_name,
...
                  group_paired_ends=False, compress_inputs=False,
-                 output_type=OutputType.FLAT_TEXT, output_is_directory=False,
+                 output_type=OutputType.FLAT_TEXT, output_is_directory=True,
                  output_names=None):
         self.tool_name = tool_name
         self.tool_version = tool_version
         self.tool_args = tool_args
         self.output_name = output_name
-        self.output_path = output_path
+        self.output_path = os.path.abspath(output_path)
```
lebovic, line 44: @bcai2 fyi this broke not specifying an output path. fixing in the async PR + adding an integration test to catch this case in the future

PR #99, opened 2022-02-10
feat: custom S3 database support

bcai2:
Adds:
- Support for custom databases, specified by the `custom_database_path` argument for `toolchest.kraken2()`. (Custom DBs are only enabled for `kraken2`.)
  - A custom database path must be given as an AWS S3 URI that must be accessible from Toolchest.
- Integration test for custom databases.

Modifies:
- The `test_kraken2_s3` integration test no longer unnecessarily downloads the input to local storage.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "toolchest-client"
-version = "0.8.1"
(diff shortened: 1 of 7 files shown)


lebovic (approved):
LGTM else the failing test / required arg

PR #100, opened 2022-02-10
v0.8.2

bcai2:
Adds:
- custom S3 database support (#99)
- add support for outputting to an S3 bucket (#96)


toolchest_client/tools/tool.py
```diff
@@ -217,7 +222,7 @@ def _preflight(self):
 
         # Check if the given output_path is a directory, if required by the tool
         # and if the user provides output_path.
-        if self.output_path:
+        if self._is_local_path():
```
lebovic, line 225: @jherr-dev nit: `_is_local_output_path()` would have been more clear in this context

lebovic: although I think what you added with the function is still a net improvement

PR #101, opened 2022-02-15
fix: set default version number

bcai2:
Sets the version number to a default instead of erroring if the client is run from source (i.e., without the `toolchest-client` package being installed via pip).

Open question: the version number defaults to `0.0.0`, which can be confusing -- are there any other labels that might be better (e.g., `dev` or just the empty string)?


lebovic (approved):
That's fine, why not `__version__` of `None`?

bcai2:
Reasonable, let's go with that

PR #103, opened 2022-02-16
feat: support async execution

lebovic:
Adds:
- Support for async execution

See https://gist.github.com/lebovic/72fbb857119f1667c7959a4d7e28cd50 (or the integration test) for a hacky example on how to run Toolchest with async execution.


tests/test_async.py
```diff
@@ -0,0 +1,45 @@
...
+    toolchest_run = toolchest.kraken2(
+        read_one="s3://toolchest-integration-tests/synthetic_bacteroides_reads.fasta",
+        custom_database_path=custom_db,
+        output_path=output_dir_path,
+        is_async=True,
+    )
+
+    run_status = ''
+    while run_status != toolchest.Status.READY_TO_TRANSFER_TO_CLIENT:
+        time.sleep(5)
```
jherr-dev, line 34: Have you checked average run time for the test? 5 seconds seems non representative of actual use and even a 30 second sleep wouldn't slow down integration tests too much

toolchest_client/tools/tool.py
```diff
@@ -387,6 +397,9 @@ def run(self):
             and (self.group_paired_ends or self.num_input_files == 1) \
             and self._system_supports_parallel_execution()
 
+        if should_run_in_parallel:
+            self.is_async = False
```
jherr-dev, line 401: We may want to print a warning here in case someone is expecting an async parallel run.

toolchest_client/tools/api.py
```diff
@@ -8,7 +8,7 @@
     STARInstance, Test, Unicycler
 
 
-def bowtie2(inputs, output_path=None, database_name="GRCh38_noalt_as", database_version="1", tool_args=""):
+def bowtie2(inputs, output_path=None, database_name="GRCh38_noalt_as", database_version="1", tool_args="", **kwargs):
```
jherr-dev, line 11: Is the `**kwargs` just to avoid adding `is_async` (and future new args) to the subclasses? I'm not really a fan of that from a readability perspective.

tests/test_async.py
lebovic: The runtime for this test is 1:15 at most :)

Not worried about the wasted cycles here

toolchest_client/tools/tool.py
lebovic: That's a good idea, will add

tests/test_async.py
jherr-dev: Yeah that's short enough that 5 seconds would a representative time anyways.

toolchest_client/api/status.py
```diff
@@ -2,12 +2,29 @@
 toolchest_client.api.status
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-This module contains a Status class for status updates to the Toolchest API.
+This module contains a function to check pipeline_segment_instance statues and status enums.
```
bcai2, line 5: Typo, statuses

toolchest_client/tools/api.py
lebovic: As I see it, there are two categories of args:

1) Tool specific args that change from tool to tool, or at least handled differently in different tool cases (e.g. inputs could also be read_one + read_two, etc).

2) Things that apply to _all_ tools, and aren't interface specific.

I think all args for case 1 should be exposed at a top level (e.g. inputs), and the args for case 2 should be contained within kwargs. I would argue that `parallelize` and other similar args also fall into case 2, but we haven't really generalized that beyond STAR yet.

afaik `**kwargs` should be used when you want to pass handling of an undetermined amount of args down to a lower level. I agree that it's not usually the best use case, but I think in this case (passing in api.py -> tool specific implementation  (e.g. kraken2.py) -> tool.py) it makes it a lot easier.

lebovic: (the undetermined args being the other `is_async`-like args that we enable at the tool level after this point

bcai2 (approved):
LGTM! a couple typos, and I am mildly concerned about `Query` being circularly imported, but it seems justifiable to me over having a separate module just for checking the status

bcai2:
Also let's hold off on merging/running integration tests until the bugfix is merged

toolchest_client/tools/api.py
jherr-dev: I see the two different argument types but I think this makes the ux generally worse. If I'm a new user of the client I probably have 1 or 2 tools I want to use that I may already be familiar with the tool args but not the toolchest args. So if I want to write `toolchest.kraken2(...)` in an ide and see the args list from there it'll list some args but then hit `**kwargs`and now the user needs to explore tool.py to find the full list. This is fine for developers but since the primary idea is to make it straightforward for bioinformaticians and other people who aren't pure developers this hurts that goal (and even as a developer not having to look up exact arg names is nice).

bcai2: As long as what can be added in `**kwargs` is documented eventually, I'm okay with this. There's another parameter (`skip_decompression`) that I want to add, but it's generalizable much like `is_async` is in this case.

bcai2: I do see the argument of adding it for usability + visibility that @jherr-dev suggested, though -- if that's our goal, then we should optimize for that now and refactor later if argument glut hinders readability and development speed.

lebovic: I agree with the argument for making this easier for bioinformaticians / etc, but this functionality – and most generalizable functionality that I'll imagine we're going to add – aren't going to fall into that use case.

People who are using `is_async` (and I would imagine `skip_decompression`) are much more competent.

Keep in mind that the specific user we're building for in this case was also okay building directly into our completely non-documented API.

lebovic: I do agree that we may eventually wan't to refactor to focus on ease of use via an IDE / etc / but I think that's premature optimization for UX in this case, at the cost of development speed.

I don't feel _too_ strongly about this and am okay with refactoring. Will ping Bryce to discuss live.

toolchest_client/api/download.py
```diff
@@ -23,34 +23,38 @@
...
 
-def download(output_path, s3_uri=None, pipeline_segment_instance_id=None,
+def download(output_path, s3_uri=None, pipeline_segment_instance_id=None, run_id=None,
              output_file_keys=None, skip_decompression=False, output_type=None):
     """Downloads output to `output_path`.
 
     One of `s3_uri`, `pipeline_segment_instance_id`, or `output_file_keys` must
     be provided. If `output_file_keys` is omitted, this function will attempt to
-    extract access keys from `s3_uri` or `pipeline_segment_instance_id` via
+    extract access keys from `s3_uri` or `id` via
```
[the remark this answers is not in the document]
lebovic: Whoops, thanks! Renaming to `run_id`, because nobody knows what a `pipeline_segment_instance_id` is :)

toolchest_client/api/status.py
lebovic: Or was it?

Screen Shot 2022-02-15 at 9 33 00 PM

lebovic: Jk, will fix

toolchest_client/tools/api.py
lebovic: Resolved live with Bryce

PR #108, opened 2022-02-18
feat: add full alphafold functionality

jherr-dev:
- Enables all AlphaFold whitelisted parameters (reduced db and prokaryote list implemented now)
- Refactors code to match other tool's arg processing.
- Updates doc string


bcai2 (approved):
LGTM! pending a few nits and local testing (just to make sure that expression of string concatenation in tool_args works)

PR #109, opened 2022-02-23
refactor: S3 download from fully private bucket

bcai2:


tests/util/s3.py
```diff
@@ -1,7 +1,15 @@
...
-    s3 = boto3.client('s3')
-    s3.download_file('toolchest-integration-tests', s3_file_key, output_file_path)
+    s3_client = boto3.client('s3')
+    try:
+        s3_client.download_file('toolchest-integration-tests', s3_file_key, output_file_path)
+    except ClientError:
+        try:
+            s3_client.download_file('toolchest-integration-tests-private', s3_file_key, output_file_path)
+        except ClientError as err:
+            raise err
```
lebovic, line 15: This is a clever way of trying normal vs. private integration test bucket, but it's pretty ambiguous and not a common pattern.

What about an `is_private` argument that defaults to `False` that can be set manually for the files that are in the private bucket? (or similar)

PR #114, opened 2022-03-01
fix: remove pysam due to multiple issues

jherr-dev:
Pysam has caused multiple issues as a package and STAR parallelization is not currently used so this pr fully removes pysam as a dependency. Either a different library or custom sam file merging code is planned to be implemented later so parallelization framework is remaining in the code for now.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -27,7 +27,6 @@ boto3 = "^1.18.29"
 python = "^3.6"
 requests = "^2.25.1"
 python-dotenv = "^0.18.0"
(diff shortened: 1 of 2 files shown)


bcai2 (approved):

lebovic (changes requested):
Can you also mark STAR as disabled for parallelization?

jherr-dev:
> Can you also mark STAR as disabled for parallelization?

Already disabled in the tool definition which is why this pr doesn't change that.

PR #118, opened 2022-03-02
Add clustalo support

lebovic:


toolchest_client/api/download.py
```diff
@@ -61,6 +61,10 @@ def download(output_path, s3_uri=None, pipeline_segment_instance_id=None, run_id
             error_message = "Details of files to download were not provided."
             raise ToolchestDownloadError(error_message) from None
 
+    # If the output path is /path/sample.fastq but the run also has tarred log files, we want to unpack at the directory
+    if output_file_keys["primary_name"]:
+        output_path = os.path.dirname(output_path)
+
```
lebovic, line 67: I feel like this could break somehow (maybe by direct download?), but I can't think of a reason why this shouldn't work right now.

That said, the name `primary_name` is kinda ambiguous. Can't think of a better one right now, but open to changing this.

PR #124, opened 2022-03-07
feat: add demucs

lebovic:


toolchest_client/tools/demucs.py
```diff
@@ -0,0 +1,30 @@
...
+
+This is the Demucs implementation of the Tool class.
+"""
+from . import Tool
+from toolchest_client.files import OutputType
+
+
+class Demucs(Tool):
+    """
+    The Demucs Omega implementation of the Tool class.
```
bcai2, line 13: typo: is it Demucs Omega or just Demucs?

toolchest_client/tools/tool.py
```diff
@@ -462,7 +462,7 @@ def run(self):
             print(
                 "\nToolchest run failed. "
                 "For support, contact Toolchest with the error log (above) and the following details:\n\n"
-                f"run_id: \n{pretty_print_run_ids}\n"
+                f"run_id: {pretty_print_run_ids}\n"
```
bcai2, line 465: The extra line is in there in case there are multiple run IDs (for multiple threads), but if you think it's best, I'm okay with changing this for now since we disabled parallelization

bcai2 (approved):
couple nits but otherwise LGTM 👍

toolchest_client/tools/demucs.py
lebovic: That is indeed a typo

toolchest_client/tools/tool.py
lebovic: It looks very awkward when one run fails – there's a newline, and the one id is partially indented.

Keeping out since it's okay with you.

PR #133, opened 2022-03-31
chore: increase # of CircleCI parallel runs

bcai2:
Increases # of CircleCI parallel runs from 2 to 10.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -33,7 +33,7 @@ jobs:
   integration-tests:
     executor: python/default
     working_directory: ~/repo
-    parallelism: 2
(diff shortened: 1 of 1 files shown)


lebovic:
@bcai2 did this end up working? It looks like the parallelism is still two on CircleCI

jherr-dev:
> @bcai2 did this end up working? It looks like the parallelism is still two on CircleCI

it worked in that it increased the number of instances however with env setup on each independently it didn't really save any time and was reverted. could try again with like 4-6 instances to not have as much of a setup bottle neck and be potentially faster than 2

PR #135, opened 2022-04-01
fix: catch generic http error

bcai2:
Catches `Exception` when checking whether inputs are HTTP URLs.

(See comment.)


toolchest_client/files/http.py
```diff
@@ -29,7 +29,7 @@ def path_is_http_url(path):
     """
     try:
         get_http_url_file_size(get_url_with_protocol(path))
-    except (InvalidURL, HTTPError, InvalidSchema, LocationParseError):
+    except (InvalidURL, HTTPError, InvalidSchema, LocationParseError, UnicodeError, Exception):
```
bcai2, line 32: This would be cleanest if this was just `except Exception:`, but PEP8 style / IDEs don't like how `except Exception` is too broad. This feels like a compromise of a workaround.

jherr-dev (approved):

PR #137, opened 2022-04-07
feat: retry on status check timeout

bcai2:
Retries the status check in `api/query.py` when waiting for a job to complete, up to a retry limit (currently hardcoded to 5 for each new `Query`).


lebovic (approved):
LGTM as an immediate fix. We should have timeouts / retries enabled across the board though – a generic version of this would probably be apt.

PR #139, opened 2022-04-13
feat: add integration test for bowtie2

jherr-dev:


lebovic (approved):
It's better than what we have, so LGTM!

That said, I think a different file (e.g. human, non-microbial) would better test Bowtie 2

bcai2 (approved):
LGTM, echoing comment in other review

PR #142, opened 2022-04-13
fix: switch hash to int from string

jherr-dev:


tests/test_diamond.py
```diff
@@ -8,7 +8,8 @@
 if toolchest_api_key:
     toolchest.set_key(toolchest_api_key)
 
-DEFAULT_OUTPUT_HASH = 952562472
+DEFAULT_BLASTP_OUTPUT_HASH = 952562472
+DEFAULT_BLASTX_OUTPUT_HASH = 883070112
```
lebovic, line 12: Why make this a global?

lebovic (approved):

bcai2 (approved):

tests/test_diamond.py
jherr-dev: If we want to test more cli options (other than format) it's DRY but probably is a bit premature.

PR #145, opened 2022-04-14
feat: add unicycler integration test

jherr-dev:


tests/test_unicycler.py
```diff
@@ -0,0 +1,30 @@
...
+    output_dir_path = f"./{test_dir}/"
+
+    toolchest.unicycler(
+        output_path=output_dir_path,
+        read_one="s3://toolchest-integration-tests/r1.fastq.gz",
+        read_two="s3://toolchest-integration-tests/r2.fastq.gz",
+        long_reads="s3://toolchest-integration-tests/long_reads.fasta.gz"
+    )
+
+    assert hash.unordered(f"{output_dir_path}assembly.fasta") == DEFAULT_OUTPUT_HASH
```
bcai2, line 30: IMO it's cleaner just to put the value here instead of making it a variable. If we want to prevent "magic numbers", then just moving it into the scope of the function (vs. making it a global) is cleaner as well

bcai2 (approved):
LGTM if works, see comment

PR #154, opened 2022-04-29
feat: custom API URL support

bcai2:
Adds:
* Support for custom API URLs, via `toolchest.get_api_url()` and `toolchest.set_api_url()`. Custom API URLs can be set through the `TOOLCHEST_API_URL` environment variable.

Modifies:
* URL handling throughout the package. Instead of global variables stored in `api/urls.py`, every URL is now retrieved through an encapsulated function.

### DIFF
--- a/toolchest_client/api/urls.py
+++ b/toolchest_client/api/urls.py
@@ -9,11 +9,11 @@ Toolchest queries and API calls.
 import os
 
 # Base URLs used by the server API.
-BASE_URL = os.environ.get("BASE_URL", "https://api.toolche.st")
+API_URL = os.environ.get("API_URL", "https://api.toolche.st")
 
 PIPELINE_SEGMENT_INSTANCES_ROUTE = "/pipeline-segment-instances"
-PIPELINE_SEGMENT_INSTANCES_URL = BASE_URL + PIPELINE_SEGMENT_INSTANCES_ROUTE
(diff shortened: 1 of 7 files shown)


toolchest_client/api/urls.py
```diff
@@ -8,12 +8,41 @@
...
+    Usage::
+
+        >>> import toolchest_client as toolchest
+        >>> toolchest.set_api_url("http://your.custom.api.url.here")
+
+    """
+    if custom_api_url:
+        os.environ["TOOLCHEST_API_URL"] = custom_api_url
+    else:
+        os.environ.pop("TOOLCHEST_API_URL", None)
```
lebovic, line 48: What's the intent here?

lebovic (approved):

toolchest_client/api/urls.py
lebovic: Wouldn't the same be accomplished by the other block of the if/else?

bcai2: Not exactly, since the key `TOOLCHEST_API_URL` would still exist in `os.environ` if something like `pop` wasn't called. (In fact, for any `os.environ["variable_name"] = x` call, `x` has to be a string, or Python will error.)

The `pop` call clears the env var key to avoid all future calls of `get_api_url` returning something like `None`. (It does feel inelegant, but there aren't any better options, according to StackOverflow.)

lebovic: Gotcha, makes sense!

PR #180, opened 2022-06-03
feat: add upload and download progress indicators

jherr-dev:


bcai2 (approved):
LGTM, we might have to iterate on output format but this is definitely a step up.

bcai2 (approved):

lebovic (approved):
LGTM, although:

1) Base branch is `main` right now
2) In the future, we should use something other than raw bytes

PR #181, opened 2022-06-06
fix: standardize output path

bcai2:
* Refactors `output_path` to refer _only_ to a directory. If a tool needs a file name specified as well, it should be passed into the `output_primary_name` parameter.
* Removes redundant output variables (`output_name`) and all logic related to whether `output_path` is a directory.
* Adds the `output_path` parameter in `Output.download()`. This should be used in place of the `output_dir` parameter, though the `output_dir` parameter is still supported for backwards compatibility.

This is a **breaking change**. The function calls for the following tools are affected:
* `clustalo`
* `diamond_blastp`
* `diamond_blastx`
* `rapsearch2`

For these tools, `output_path` formerly referred to an output **file**. The base file name should now be parsed into `output_primary_name` instead, and `output_path` should contain the directory the output file will be in.

### DIFF
--- a/toolchest_client/tools/tool.py
+++ b/toolchest_client/tools/tool.py
@@ -206,9 +206,8 @@ class Tool:
 
     def _warn_if_outputs_exist(self):
         """Warns if default output files already exist in the output directory"""
-        output_dir = self.output_path if self.output_is_directory else os.path.dirname(self.output_path)
         for file_path in self.expected_output_file_names + ["output", "output.tar.gz"]:
-            joined_file_path = os.path.join(output_dir, file_path)
+            joined_file_path = os.path.join(self.output_path, file_path)
             if os.path.exists(joined_file_path):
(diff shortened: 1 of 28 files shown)


toolchest_client/tools/tool.py
```diff
@@ -297,7 +279,6 @@ def _handle_termination(self, signal_number, *_):
         if self.terminating:
             raise InterruptedError("Toolchest client force killed")
         self.terminating = True
-        sentry_sdk.set_tag('send_page', False)
```
lebovic, line 300: Is this removal intentional?

lebovic (approved):
LGTM

toolchest_client/tools/tool.py
bcai2: Not as part of this PR. I'm guessing that this resulted from a merge from staging, but I'll investigate further

bcai2: Looks like this was from a merge commit from staging into this branch.

PR #183, opened 2022-06-10
feat: add a cli for kraken2 and test

jherr-dev:
Adds a cli so that kraken2  can be run by installing the cli and then adding `toolchest` before an existing command. It will attempt to heal tool versions and databases. This will be the base for future tools to be used on the cli.


toolchest_client/cli/cli.py
```diff
@@ -0,0 +1,17 @@
+import sentry_sdk
```
lebovic, line 1: The Sentry SDK is not longer in the project, this + the tag below should be removed

lebovic: @jherr-dev before I continuing reviewing, was this tested? Could have passed if you have `sentry_sdk` installed elsewhere

lebovic (approved):
We'll want to improve on this once we revisit the CLI, but LGTM for a first step!

PR #189, opened 2022-06-22
fix: remove status msg if run failed

bcai2:
Removes the last printed status message (between the error output and the "job failed" message) if a job did not finish successfully. For example:

```
<rest of error message>
Error: The sequences are expected to be proteins but only contain DNA letters.

Running 1 job | Duration: 0:01:05 | 1 job executing  // this line gets removed by this PR

Toolchest run failed. For support, contact Toolchest with the error log (above) and the following details:

run_id:         "example-run-id"
```


jherr-dev (approved):

lebovic (approved):
👍

PR #193, opened 2022-06-26
fix: handle pipeline segments with no DB

bcai2:
* Handles errors that occur during uploading (i.e. before a query's `ThreadStatus` changes to `UPLOADED`). This fixes the infinite hanging bug.
* Avoids erroring if the requested tool does not specify a database.

### DIFF
--- a/toolchest_client/tools/tool.py
+++ b/toolchest_client/tools/tool.py
@@ -298,7 +298,7 @@ class Tool:
             if thread.is_alive() and thread_status not in {ThreadStatus.COMPLETE, ThreadStatus.FAILED}:
                 self.query_thread_statuses[thread_name] = ThreadStatus.INTERRUPTING
 
-        self._wait_for_threads_to_finish(check_health=False)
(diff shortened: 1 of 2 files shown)


toolchest_client/tools/tool.py
```diff
@@ -342,6 +334,24 @@ def _wait_for_threads_to_finish(self, check_health=True):
         for thread in self.query_threads:
             thread.join()
 
+    def _wait_for_threads_to_upload(self):
+        """Waits for all jobs to finish uploading. To be used only at the start of a run."""
+        while True:
```
lebovic, line 339: `while True` is almost always anti-pattern – works, but worth flagging

lebovic (approved):
👍

toolchest_client/tools/tool.py
bcai2: Fair, that's what I had thought -- we can refactor this, I just wanted to avoid the extra `time.sleep(5)` call, which can be noticeable if the file is small

PR #194, opened 2022-06-26
refactor: upload loop

bcai2:
* Refactors the `_wait_for_threads_to_upload` loop to avoid `while True` syntax.


lebovic (commented):
I might be misreading, but is `uploading` ever not `True`?

bcai2:
`uploading` is set to `False` when none of the remaining threads have statuses `INITIALIZING`, `INITIALIZED`, or `UPLOADING` (i.e., they've all moved past the `UPLOADING` stage). See lines 348-351 (which might be obscured in the change preview).

jherr-dev (approved):

PR #196, opened 2022-06-28
fix: async success in _check_thread_health

bcai2:
Updates the expected "success" status in `_check_thread_health()` to support async runs.


lebovic (approved):
👍

PR #197, opened 2022-06-30
feat: add python3 tool implementation

jherr-dev:
Implements a tool to run python3 scripts via Toolchest. Script must use the expected local input and output directories but are otherwise not limited.

### DIFF
--- a/toolchest_client/tools/api.py
+++ b/toolchest_client/tools/api.py
@@ -7,7 +7,6 @@ This module contains the API for using Toolchest tools.
 import os.path
 from datetime import date
 
-from toolchest_client import ToolchestException
 from toolchest_client.tools import AlphaFold, Bowtie2, CellRangerCount, ClustalO, Demucs, DiamondBlastp, DiamondBlastx,\
     Kraken2, Megahit, Python3, Rapsearch2, Shi7, ShogunAlign, ShogunFilter, STARInstance, Test, Unicycler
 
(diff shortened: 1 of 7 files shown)


toolchest_client/tools/api.py
```diff
@@ -388,6 +391,44 @@ def megahit(output_path=None, tool_args="", read_one=None, read_two=None, interl
     return output
 
 
+def python3(script, inputs=[], output_path=None, tool_args="", **kwargs):
+    """Runs Python3 via Toolchest.
+    NOTE: This tool is currently only available for customers with contracts
```
lebovic, line 396: Rephase: e.g. "this is a restricted tool, you can request access at ___"

toolchest_client/tools/api.py
```diff
@@ -388,6 +391,44 @@ def megahit(output_path=None, tool_args="", read_one=None, read_two=None, interl
...
+    The script has a few requirements if inputs or outputs are expected. Inputs will be uploaded to `./input/` so the
+    script must use that prefix with hardcoded paths or when adding to tool_args to be parsed via sys.argv. Any output
+    that is expected to be returned needs to be stored at `./output/`. The directory will be compressed via pigz and
+    decompressed on the client.
```
lebovic, line 400: This can be rephrased to be more digestible

toolchest_client/tools/api.py
```diff
@@ -388,6 +391,44 @@ def megahit(output_path=None, tool_args="", read_one=None, read_two=None, interl
...
+
+    :param script: Path to the python script to run
+    :param inputs: (optional) Path(s) to the input files to run with the python script. Will be uploaded to a directory
+    `./input/` for use in the on the instance.
+    :param output_path: (optional) Local path to where the output file(s) will be downloaded.
+    :param tool_args: (optional) Additional arguments to be passed to Python3. Should NOT specify the script here but
+    any other args after the script should be included.
```
lebovic, line 407: Rephrase with a positive example and/or a more clear example of what not to do, or just remove the added context.

(More specifically, modify or remove "Should NOT specify the script here but any other args after the script should be included.")

toolchest_client/tools/api.py
lebovic: "The script has a few requirements if inputs or outputs are expected." – not necessary

"Inputs will be uploaded to `./input/` so the script must use that prefix with hardcoded paths or when adding to tool_args to be parsed via sys.argv" - hard to understand

"Any output that is expected to be returned needs to be stored at `./output/`" - make more clear what this means, what needs to store what at ./output/?

"The directory will be compressed via pigz and decompressed on the client." - more detail than is necessary, but still lacks implication (maybe "The output directory is compressed and returned, meaning chaining Python 3 calls via Toolchest is not possible")

toolchest_client/tools/api.py
```diff
@@ -388,6 +391,44 @@ def megahit(output_path=None, tool_args="", read_one=None, read_two=None, interl
...
+        >>> toolchest.python3(
+        ...     script="./path/to/script.py",
+        ...     output_path="./path/to/output/base",  # outputs
+        ...     tool_args="",
+        ... )
+    """
+    if type(inputs) is str:
+        inputs = [inputs]
+    elif type(inputs) is not list:
+        raise ToolchestException('Unknown input type')
```
lebovic, line 419: Why was this added, did you hit an error? I'm curious because this seems like a new pattern

toolchest_client/tools/python3.py
```diff
@@ -0,0 +1,28 @@
...
+    """
+    def __init__(self, tool_args, inputs, output_path, **kwargs):
+        super().__init__(
+            tool_name="python3",
+            tool_version="3.9.1",
+            tool_args=tool_args,
+            output_path=output_path,
+            inputs=inputs,
+            min_inputs=1,
+            max_inputs=100,
```
lebovic, line 24: This is default behavior, safe to remove?

toolchest_client/tools/python3.py
```diff
@@ -0,0 +1,28 @@
...
+            parallel_enabled=False,
```
lebovic, line 25: Same here

toolchest_client/tools/tool_args.py
```diff
@@ -385,6 +385,11 @@
             "--min-contig-len": 1,
         },
     },
+    "python3": {
```
lebovic, line 388: Might be worth blacklisting some args – e.g. `-i` for spawning an interactive Python environment

lebovic (changes requested):
Overall looks good, comments center around documentation

toolchest_client/tools/api.py
jherr-dev: Just being overly cautious. I'll remove it

lebovic (approved):

PR #202, opened 2022-07-07
feat: output tool name/version

bcai2:
Adds:
* Tool name/version in the the output object returned from tool function calls.
* Integration test for verifying the output object's variables.

Modifies:
* `Output.output_path`, refactored to be the output directory (as per #181). All paths to output files can be found in `Output.output_file_paths`.


lebovic (approved):
👍

PR #207, opened 2022-07-15
feat: live output streaming

bcai2:
Adds basic live output streaming. Prints lines outputted by the tool to the terminal, all while maintaining the job status line at the bottom.

Some snapshots of the bottom lines of the same job:
```
Running 1 job | Duration: 0:00:10 | 1 job executing
```

```
Searching alignments...  [0.205s]
Processing query block 1, reference block 1/1, shape 1/2, index chunk 4/4.
Building reference seed array...  [0.032s]
Building query seed array...  [0.032s]
Indexed query seeds = 607268/2611087 (23.26%), reference seeds = 607268/2611087 (23.26%)
Soft masked letters = 1338/2611087 (0.05%), 1338/2611087 (0.05%)
Computing hash join...  [0.015s]
Running 1 job | Duration: 0:00:55 | 1 job executing
```

```
Time (Load seed hits)        = 0.00586s (wall)
Time (Sort seed hits)        = 0.003453s (wall)
Time (Extension)             = 1.85312s (wall)
Temporary disk space used (search): 0.00109231 GB
Reported 28519 pairwise alignments, 28519 HSPs.
14323 queries aligned.
output.tar.gz  496.2KB of 496.2KB (100.00%)
Running 1 job | Duration: 0:01:05 | 1 job complete
```

Notes:
- New output stream lines are fetched with every job status check, which occurs every 5 seconds or so, so the output may appear "staggered". For "live" streaming with a faster response time, a server-to-client stream will need to be built + maintained.
- Lines may be occasionally dropped by the backend. No errors are thrown by either the backend or the client if this happens.

### DIFF
--- a/toolchest_client/api/query.py
+++ b/toolchest_client/api/query.py
@@ -36,7 +36,7 @@ class Query:
 
     # Period (in seconds) between requests when waiting for job(s) to finish executing.
     WAIT_FOR_JOB_DELAY = 1
-    # Multiple of seconds used when pretty printing job status to output.
+    # Multiple of seconds used when pretty printing job status + tool output stream to output.
     PRINTED_TIME_INTERVAL = 5
     # Max number of retries on status check timeouts.
     RETRY_STATUS_CHECK_LIMIT = 5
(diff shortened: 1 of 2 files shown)


toolchest_client/api/query.py
```diff
@@ -55,12 +55,18 @@ def __init__(self, stored_output=None, is_async=False, pipeline_segment_instance
                 self.PIPELINE_SEGMENT_INSTANCE_URL,
                 "status",
             ])
+            self.OUTPUT_STREAM_URL = "/".join([
+                self.PIPELINE_SEGMENT_INSTANCE_URL,
+                "output-stream",
+            ])
```
lebovic, line 61: These URLs are getting a bit unwieldy, and `OUTPUT_STREAM_URL` is defined three times and only used once.

Cleanup is outside the scope of this PR, but as an intermediate step maybe move this to `get_output_stream_latest_lines`?

PR #211, opened 2022-07-21
feat: scaffolding for DIAMOND BLASTX parallelization

lebovic:


toolchest_client/files/unpack.py
```diff
@@ -34,10 +33,6 @@ def unpack_files(file_path_to_unpack, is_compressed):
         unpacked_paths = ["/".join([unpacked_outputs_dir, file_name]) for file_name in unpacked_file_names]
         unpacked_file_paths = [os.path.normpath(path) for path in unpacked_paths if os.path.isfile(path)]
 
-        # If only 1 file is unpacked, just return path instead of [path].
-        # This is to be consistent with the return value from the other output types.
-        if len(unpacked_file_paths) == 1:
-            return unpacked_file_paths[0]
```
lebovic, line 40: This is an interface change: when a `.tar.gz` is returned, `output_file_paths` was previously:
- a list if there were multiple files present
- a single string if there was only one file present

That's suboptimal – imagine a situation in which you want to `[] += output.output_file_paths`. A string would return `['f', 'i', 'l', 'e']`, and a list would return `['file']`.

I hit this issue on the infra-side, so adjusted it in this PR

lebovic: I don't think this is a breaking change, because I couldn't find any tools that are compressing output which don't have multiple files at this point

toolchest_client/tools/diamond.py
```diff
@@ -33,18 +33,18 @@ class DiamondBlastx(Tool):
...
             tool_version="2.0.13",
             tool_args=tool_args,
             output_primary_name=output_primary_name,
             inputs=inputs,
             database_name=database_name,
             database_version=database_version,
-            parallel_enabled=False,
             output_type=OutputType.GZ_TAR,
             output_path=output_path,
-            expected_output_file_names=[output_primary_name, "diamond.log"],
```
lebovic, line 48: `diamond.log` is demoted to a non-critical return; when distributed the individual logs are returned in `./artifacts`

toolchest_client/tools/tool.py
```diff
@@ -121,9 +121,10 @@ def _validate_tool_args(self):
         is shown.
         """
 
-        whitelist = TOOL_ARG_LISTS[self.tool_name]["whitelist"]  # all tools have a whitelist
-        dangerlist = TOOL_ARG_LISTS[self.tool_name].get("dangerlist", [])  # some tools have a dangerlist
-        blacklist = TOOL_ARG_LISTS[self.tool_name].get("blacklist", [])  # some tools have a blacklist
+        tool_category = self.tool_name.replace("_parallel", "")
```
lebovic, line 124: This is a bit hacky, not sure if there's a better way

jherr-dev (approved):
LGTM. Only nit is the mixed use of `distributed` and `parallel` which could be clearer to settle on one but I'm worried about that.

lebovic:
> LGTM. Only nit is the mixed use of `distributed` and `parallel` which could be clearer to settle on one but I'm worried about that.

Fair – `distributed` was a last minute change, because for client-side parallelization I had to clarify that "parallel" meant "multithreaded" to many people.

Internally, I'm okay with using "distributed" or "parallel".

PR #220, opened 2022-08-02
feat: support for using a docker image with python3

jherr-dev:
Allows for a Docker image (that supports Python3) to be provided as execution context for the script, to add dependencies outside of the Python Standard Library.


tests/test_python3.py
```diff
@@ -26,3 +28,32 @@ def test_python3():
...
+@pytest.mark.integration
+def test_python3_with_docker():
+    """
+    Tests adding dependencies to python3 via a custom docker image
+
+    Specifically tests matrix multiplication via numpy
+    """
+    client = docker.from_env()
+    client.images.build(
+        path="./util/",
```
lebovic, line 42: If running the test from anywhere but the directory of the test itself, this fails. For context, the current working directory is wherever pytest is invoked.

Maybe could anchor the relative location to the file itself, a-la https://github.com/trytoolchest/toolchest-client-python/blob/main/toolchest_client/files/tests/test_merge.py#L7?

lebovic (commented):
Looks good! I see one place where this can be significantly improved: error messages when Docker is not installed or running.

If Docker isn't installed or running, the error message is directly propagated from the Docker client:

```
docker.errors.DockerException: Error while fetching server API version: ('Connection aborted.', ConnectionRefusedError(61, 'Connection refused'))
```

Imo, the error messages and path forward can be much more clear. Maybe an explicit check to see if Docker is installed + the engine is running?

lebovic (approved):
👍

PR #221, opened 2022-08-03
feat: support for custom DB automation

bcai2:
Adds:
- `database_primary_name` parameter in `toolchest.add_database` and `toolchest.update_database`, for specifying which file to use as the DB if uploading multiple files
- new integration tests for DB add/update that upload a real DB and run a follow-up job with real data

Refactors:
- `update_database` and `add_database` now have the ability to set `is_async` to `False` to ensure that a DB is fully transferred before any jobs are run (it will still default to `True`)
- add/update DB function docstrings
- job completion messages, now edited to make more sense with DB update runs


tests/test_database_update.py
```diff
@@ -12,53 +13,125 @@
...
-    assert output.database_name == "standard"
-    assert output.database_version
+    # Run job on updated DB
+    toolchest.bowtie2(
+        inputs="s3://toolchest-integration-tests/bowtie2/fruitfly-ncbi/GSM868349.fastq.gz",
+        output_path=output_dir_path,
+        database_name=update_db_output.database_name,
+        database_version=update_db_output.database_version,
+    )
+    print(os.path.getsize(output_file_path))  # replace w/ file size assertion
```
lebovic, line 47: Should this be a file size assertion?

tests/test_database_update.py
```diff
@@ -12,53 +13,125 @@
...
```
lebovic, line 93: (same question here)

tests/test_database_update.py
bcai2: It should be, this is just a placeholder. I'll add actual values once we run the tests.

jherr-dev (approved):

PR #222, opened 2022-08-05
fix: skip head request for presigned s3 urls

jherr-dev:
S3 URLs don't allow head requests to get content-size prior to download

### DIFF
--- a/toolchest_client/api/query.py
+++ b/toolchest_client/api/query.py
@@ -217,6 +217,8 @@ class Query:
             'input-files'
         ])
         file_name = os.path.basename(input_file_path)
+        if "?" in file_name:
+            file_name = file_name.split("?")[0]
         input_is_in_s3 = path_is_s3_uri(input_file_path)
         input_is_http_url = path_is_http_url(input_file_path)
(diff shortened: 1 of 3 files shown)


bcai2 (approved):

toolchest_client/api/query.py
```diff
@@ -217,6 +217,8 @@ def _register_input_file(self, input_file_path, input_prefix, input_order):
             'input-files'
         ])
         file_name = os.path.basename(input_file_path)
+        if "?" in file_name:
+            file_name = file_name.split("?")[0]
```
lebovic, line 221: How does this handle `?`s in local file paths?

This seems specific to HTTP URLs

toolchest_client/files/public_uris.py
```diff
@@ -52,6 +52,8 @@ def get_http_url_file_size(url):
 
     :param url: An input URL.
     """
+    if all([x in url for x in ["https://", "s3", "amazonaws.com"]]):
+        return 0
```
lebovic, line 56: Discuss sync!

PR #223, opened 2022-08-08
fix: switch to GET with range for http accessiblity check

jherr-dev:


toolchest_client/api/query.py
```diff
@@ -217,7 +217,7 @@ def _register_input_file(self, input_file_path, input_prefix, input_order):
             'input-files'
         ])
         file_name = os.path.basename(input_file_path)
-        if "?" in file_name:
+        if "?" in file_name and "http" in input_file_path:
```
lebovic, line 220: Is it possible to move this a few lines down, and use `input_is_http_url`? `"http" in input_file_path` catches files outside of HTTP urls (e.g. "./fastq_1_from_ncbi_from_http.fastq")

toolchest_client/files/general.py
```diff
@@ -45,8 +45,8 @@ def check_file_size(file_path, max_size_bytes=None):
         # NOTE: If the file is already in S3, the size is checked as well to enforce an expected file size
         file_size_bytes = get_s3_file_size(file_path)
     elif path_is_http_url(file_path):
-        # Get file size via a HEAD request.
-        file_size_bytes = get_http_url_file_size(file_path)
+        # http inputs don't use s3 uploads so multipart check is not needed
```
lebovic, line 48: FWIW the multipart check is no longer relevant; by default we just use boto3.

jherr-dev: Do you think we should rip the whole function out or is that too big of a refactor for this pr?

lebovic: I wouldn't rip the whole function, but clarifying the cause of limits on the comments would be worth it – e.g. in this functions docstring and the new comment, clarifying that this just checks the limit that we set on a per-tool per-file basis

PR #229, opened 2022-08-19
feat: add capability for a user to request what instance type to run on

jherr-dev:
Adds `instance_type` and `volume_size` args to tool to allow users more flexibility in their runs. The args are top level args for python3 as the workloads for python3 are much more variable. The args are accessible via `**kwargs` for all other tools except AlphaFold, Demucs, and Transfer, which do not allow changing instance type.

### DIFF
--- a/toolchest_client/tools/api.py
+++ b/toolchest_client/tools/api.py
@@ -476,7 +476,7 @@ def megahit(output_path=None, tool_args="", read_one=None, read_two=None, interl
 
 
 def python3(script, inputs=None, output_path=None, tool_args="", custom_docker_image_id=None,
-            instance_type=InstanceType.COMPUTE_32, volume_size=8, **kwargs):
+            instance_type=InstanceType.COMPUTE_2, volume_size=8, **kwargs):
     """Runs Python via Toolchest. This a restricted tool, running it requires you to request access.
 
     Within your Python3 script, input files are available at `./input/`.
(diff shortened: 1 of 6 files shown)


toolchest_client/tools/api.py
```diff
@@ -216,6 +219,8 @@ def demucs(inputs, output_path=None, tool_args="", **kwargs):
         ... )
 
     """
+    if 'instance_type' in kwargs:
+        raise ToolchestException("Argument 'instance_type' is not supported by Demucs currently.")
```
lebovic, line 223: Nit: `Argument 'instance_type' is not supported` is all that's needed!

toolchest_client/tools/api.py
```diff
@@ -470,7 +475,8 @@ def megahit(output_path=None, tool_args="", read_one=None, read_two=None, interl
     return output
 
 
-def python3(script, inputs=[], output_path=None, tool_args="", custom_docker_image_id=None, **kwargs):
+def python3(script, inputs=None, output_path=None, tool_args="", custom_docker_image_id=None,
+            instance_type=InstanceType.COMPUTE_32, volume_size=8, **kwargs):
```
lebovic, line 479: We may want to default to a smaller instance class; this would exhaust the free tier pretty quickly

bcai2 (approved):

PR #235, opened 2022-08-29
feat: add salmon

lebovic:
Added:
- Salmon
- Docs for Salmon: https://docs.trytoolchest.com/docs/salmon


toolchest_client/tools/api.py
```diff
@@ -628,6 +628,76 @@ def rapsearch2(inputs, output_path=None, output_primary_name="output", database_
...
+                input_prefix_mapping[input_file] = {
+                    "prefix": tag,
+                    "order": index,
+                }
+        elif isinstance(param, str):
+            input_list.append(param)
+            input_prefix_mapping[param] = {
+                "prefix": tag,
+                "order": 0,
+            }
```
lebovic, line 686: This is nearly identical to what we do with `megahit`, goal is to preserve tool arg ordering

PR #239, opened 2022-09-01
feat: allow status checks to return the error via kwargs

jherr-dev:


lebovic (approved):
LGTM, pending it working

PR #243, opened 2022-09-02
Add HUMAnN rename mode

lebovic:


bcai2 (commented):
IMO, it would be better to specify `names` as an explicit function argument with client-side translation, a la other humann3 mode-specific parameters (`input_pathways`, for example). 

We can do this in a follow-up PR since it shouldn't involve any functionality change.

PR #245, opened 2022-09-07
feat: add fastqc tool

jherr-dev:


bcai2 (approved):
LGTM, though we'll probably want to add a test case to catch some of those prefixed args (once we have test data)

toolchest_client/tools/fastqc.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+class FastQC(Tool):
+    """
+    The FastQC implementation of the Tool class.
+    """
+    def __init__(self, tool_args, inputs, output_path, **kwargs):
+        super().__init__(
+            tool_name="fastqc",
+            tool_version="0.11.9",  # todo: allow salmon version to be set by the user
```
lebovic, line 20: Nit: `salmon` -> `fastqc`. Although we can probably also start dropping these comments / setting a default tool_version but adding as a kwarg

lebovic (approved):
:+1:

PR #248, opened 2022-09-09
fix: filter date out of fastqc output for hashing

jherr-dev:
Removes the date from FastQC html output via regex as that is the only non-deterministic part of the html.

### DIFF
--- a/tests/test_fastqc.py
+++ b/tests/test_fastqc.py
@@ -1,7 +1,7 @@
 import os
 import pytest
 
-from tests.util import hash
+from tests.util import hash, filter_output
 import toolchest_client as toolchest
 
 toolchest_api_key = os.environ.get("TOOLCHEST_API_KEY")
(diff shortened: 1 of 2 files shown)


tests/test_fastqc.py
```diff
@@ -19,12 +19,17 @@ def test_fastqc():
...
     )
 
-    assert hash.unordered(os.path.join(test_dir, "sample_r1_shortened_fastqc.html")) == 1123268405
-
-    assert os.path.getsize(os.path.join(test_dir, "sample_r1_shortened_fastqc.zip")) == 326078
+    filter_output.filter_regex(
+        os.path.join(test_dir, "sample_r1_shortened_fastqc.html"),
+        filtered_html_output_path,
+        search_regex='id="header_filename">([\\w\\s]+)<br',
+        replacement_str='id="header_filename">)<br'
```
bcai2, line 32: Correct me if I'm wrong, but doesn't this drop the left paren `(` as well as the date? (I'm presuming the date is outputted in between the parens.) 

It shouldn't matter and doesn't change the functionality of the test, so I'm fine leaving this as is.

bcai2 (approved):
LGTM, one comment

tests/test_fastqc.py
jherr-dev: The date is actually in between the html tags. The paren was missed when I was deleting the regex group from the search string to replace with the same html minus the group. I'll push a new version up with that fixed.

PR #254, opened 2022-09-13
feat: S3 prefix input handling for custom DB updates

bcai2:
* Adds an integration test for a custom database update with an S3 prefix as input and an auto-generated `database_primary_name`.
* Modifies function call for `add_database_update` to mandate `database_primary_name`.
* Removes `database_primary_name` sanitization.
* Increases CI parallelism to 8.

### DIFF
--- a/tests/test_database_update.py
+++ b/tests/test_database_update.py
@@ -10,7 +10,7 @@ if toolchest_api_key:
     toolchest.set_key(toolchest_api_key)
 
 
-@pytest.mark.integration
+@pytest.mark.integration_part
 def test_database_update_s3():
     """
     Tests custom database update for bowtie 2 using S3 files
@@ -50,7 +50,7 @@ def test_database_update_s3():
     assert hash.unordered(filtered_output_file_path) == 107700257
 
 
-@pytest.mark.integration
+@pytest.mark.integration_part
 def test_database_update_s3_prefix():
     """
     Tests custom database update for bowtie 2 using an S3 prefix and assumed primary name
--- a/tests/test_database_update.py
+++ b/tests/test_database_update.py
@@ -51,6 +51,39 @@ def test_database_update_s3():
 
 
 @pytest.mark.integration
+def test_database_update_s3_prefix():
+    """
+        Tests custom database update for bowtie 2 using an S3 prefix
+        """
+    test_dir = "temp_test_db_update_s3_bowtie2"
+    os.makedirs(f"./{test_dir}", exist_ok=True)
+    output_dir_path = f"./{test_dir}"
+    output_file_path = f"{output_dir_path}/bowtie2_output.sam"
+    filtered_output_file_path = f"{output_dir_path}/bowtie2_output.filtered.sam"
+
+    # Update DB
+    update_db_output = toolchest.update_database(
+        database_path="s3://toolchest-public-examples-no-encryption/integration-test-db/bowtie2-fruitfly",
+        tool=toolchest.tools.Bowtie2,
+        database_name="integration_test_bowtie2_fruitfly",
+        database_primary_name="Dmel_A4_1.0",
+        is_async=False,  # to ensure DB finishes uploading before tool call
+    )
+    assert update_db_output.database_name == "integration_test_bowtie2_fruitfly"
+    assert update_db_output.database_version
+
+    # Run job on updated DB
+    toolchest.bowtie2(
+        inputs="s3://toolchest-integration-tests/bowtie2/fruitfly-ncbi/GSM868349.fastq.gz",
+        output_path=output_dir_path,
+        database_name=update_db_output.database_name,
+        database_version=update_db_output.database_version,
+    )
+    filter_output.filter_sam(output_file_path, filtered_output_file_path)
+    assert hash.unordered(filtered_output_file_path) == 107700257
+
+
+@pytest.mark.integration
 def test_database_update_local():
     """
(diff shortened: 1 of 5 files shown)


tests/test_database_update.py
```diff
@@ -50,6 +50,42 @@ def test_database_update_s3():
     assert hash.unordered(filtered_output_file_path) == 107700257
 
 
+@pytest.mark.integration
+def test_database_update_s3_prefix():
+    """
+    Tests custom database update for bowtie 2 using an S3 prefix and assumed primary name
+
+    NOTE: to test auto-generation of database_primary_name, this should be run right after
+    test_database_update_s3
```
jherr-dev, line 59: I don't love interdependent, but separate, integration tests as they can have unintended consequences with future changes.  Better way to for the dependency would be have 1 test that calls add then update. Could force that for now by creating a wrapping test but that still separates code in a way that may not be clear if only looking at the underlying tests.

toolchest_client/tools/api.py
```diff
@@ -1201,10 +1200,9 @@ def add_database(database_path, tool, database_name, database_primary_name=None,
     :param database_path: Path or list of paths (local or S3) to be passed in as inputs.
     :param tool: Toolchest tool with which you use the database (e.g. toolchest.tools.Kraken2).
     :param database_name: Name of the new database.
-    :param database_primary_name: Name or path of the file to use as the primary database file
-        (i.e., what you would pass into the command line as the database), if uploading multiple
-        files. If unspecified, assumes that the *directory* of files is what will be passed in
-        as the database.
+    :param database_primary_name: Name or path of the file/prefix that will be passed as the database,
+        just like the command line call. Use `database_primary_name=None` to pass the entire *directory*
```
jherr-dev, line 1204: "path" might be a bit confusing here since a command might take `/path/to/prefix` and we only want `prefix` passed in. Similarly I think there may be a better way to say "just like the command line call".

toolchest_client/tools/tool.py
```diff
@@ -114,13 +114,16 @@ def _prepare_inputs(self):
...
                 self.database_primary_name = os.path.basename(self.database_primary_name)
-                # If only one input file exists, use that file explicitly as the DB
-                if self.num_input_files == 1 and not self.compress_inputs:
-                    self.database_primary_name = os.path.basename(self.input_files[0])
-                # If DB name isn't a prefix for an input file, (implicitly) use directory of inputs instead as DB
-                input_basenames = [os.path.basename(input_path) for input_path in self.input_files]
-                if all([self.database_primary_name not in basename for basename in input_basenames]):
-                    self.database_primary_name = None
+                # Sanitize database_primary_name if all files are local
+                # (If there is an S3 URI, it could be a prefix, so we skip overwriting database_primary_name)
```
jherr-dev, line 118: Couldn't you have a prefix for a db name with local files?

toolchest_client/tools/api.py
```diff
@@ -1142,10 +1142,9 @@ def update_database(database_path, tool, database_name, database_primary_name=No
     :param database_path: Path or list of paths (local or S3) to be passed in as inputs.
     :param tool: Toolchest tool with which you use the database (e.g. toolchest.tools.Kraken2).
     :param database_name: Name of database to update.
-    :param database_primary_name: Name or path of the file to use as the primary database file
-        (i.e., what you would pass into the command line as the database), if uploading multiple
-        files. If unspecified, assumes that the *directory* of files is what will be passed in
-        as the database.
+    :param database_primary_name: Name or path of the file/prefix that will be passed as the database,
+        just like the command line call. Use `database_primary_name=None` to pass the entire *directory*
+        of files as the database. If left unspecified, assumes the primary name of the previous version.
```
jherr-dev, line 1147: See comment on add.

toolchest_client/tools/tool.py
bcai2: Local prefix support hasn't been implemented yet. If we pass in a local directory, it will be parsed into its individual files found along the search path (before reaching this point, IIRC).

Not sure if we want to support true "prefixes" locally in this PR or at all, which is a discussion worth having.

jherr-dev: I think if the use passes a prefix as a primary name we need to support it and not override it.

jherr-dev: Which could be true for the single input case as well.

toolchest_client/tools/api.py
```diff
@@ -1201,20 +1201,20 @@ def add_database(database_path, tool, database_name, database_primary_name=None,
     :param database_path: Path or list of paths (local or S3) to be passed in as inputs.
     :param tool: Toolchest tool with which you use the database (e.g. toolchest.tools.Kraken2).
     :param database_name: Name of the new database.
-    :param database_primary_name: Name or path of the file to use as the primary database file
-        (i.e., what you would pass into the command line as the database), if uploading multiple
-        files. If unspecified, assumes that the *directory* of files is what will be passed in
-        as the database.
+    :param database_primary_name: Base name of the file/prefix that would normally be passed in
+        to the command line call. Use `database_primary_name=None` to pass the entire *directory*
+        of files as the database.
```
jherr-dev, line 1206: I would maybe say something like "to use the directory name as the database". This phrasing to me could read as if the whole directory may not be uploaded if  `database_primary_name=None`isn't used.

jherr-dev (approved):
Looks good other than a potential change in the docstring.

PR #262, opened 2022-09-22
feat: compress local inputs on custom DB updates (TC-20)

bcai2:
- Compresses inputs when passing in a single local input path for custom DB updates, in order to preserve directory structure.
- Propagates `compress_input` argument to API during input file registration.
- Adds an integration test for this case (custom DB update, local prefix input).
- Refactors integration test function calls for custom DB updates with local inputs, for clarity.

### DIFF
--- a/tests/test_database_update.py
+++ b/tests/test_database_update.py
@@ -10,7 +10,7 @@ if toolchest_api_key:
     toolchest.set_key(toolchest_api_key)
 
 
-@pytest.mark.integration
+@pytest.mark.integration_part
 def test_database_update_s3():
     """
     Tests custom database update for bowtie 2 using S3 files
@@ -50,7 +50,7 @@ def test_database_update_s3():
     assert hash.unordered(filtered_output_file_path) == 107700257
 
 
-@pytest.mark.integration
+@pytest.mark.integration_part
 def test_database_update_s3_prefix():
     """
     Tests custom database update for bowtie 2 using an S3 prefix and assumed primary name
--- a/tests/test_database_update.py
+++ b/tests/test_database_update.py
@@ -73,16 +73,8 @@ def test_database_add_and_update_s3():
 
 
 @pytest.mark.integration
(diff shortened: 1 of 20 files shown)


tests/test_database_update.py
```diff
@@ -165,3 +166,26 @@ def run_and_test_bowtie2_on_updated_db(
...
+    downloaded_full_paths = []
+    for db_file_name in db_file_names:
+        downloaded_full_path = os.path.join(input_dir_path, db_file_name)
+        if not os.path.isfile(downloaded_full_path):
+            s3.download_integration_test_input(
+                s3_file_key=f"databases/bowtie2-fruitfly-BDGP6/{db_file_name}",
+                output_file_path=downloaded_full_path,
+            )
+        downloaded_full_paths.append(downloaded_full_path)
+    return downloaded_full_paths
```
bcai2, line 191: Note that this database does not have any subdirectories. We currently don't have a database that does all of the following:
- Has subdirectories that are actually used by the tool
- Is small enough to download locally and re-upload without taking a significant amount of time

We can add a dummy subdirectory+subfile to test functionality, but this integration test wouldn't be able to detect automatically that the subfile was transferred.

jherr-dev (approved):
LGTM pending testing and future changes.

tests/test_database_update.py
```diff
@@ -73,33 +73,34 @@ def test_database_add_and_update_s3():
...
+def test_database_update_local_list():
+    base_test_database_update_local(test_with_prefix=False)
+
+
+@pytest.mark.integration
+def test_database_update_local_prefix():
+    base_test_database_update_local(test_with_prefix=True)
+
+
+def base_test_database_update_local(test_with_prefix):
```
lebovic, line 85: Nit: this looks like a good use of `@pytest.paramaterize`: https://docs.pytest.org/en/6.2.x/parametrize.html

PR #267, opened 2022-10-03
feat: output stream handling (TCP-39)

bcai2:
- Adds a `StreamingClient` that handles receiving the output stream from the Toolchest server.
- Adds the `streaming_enabled` arg to `toolchest.python3()` and `toolchest.lug()`.
  - This is set to `True` by default for both.
- Adds output streaming during tool execution for commands where streaming is enabled. The stream is printed in place of the regular job status updates when the stream is active (after stream param initialization, stopping once the output is ready to download).
  - Streaming params are retrieved in `query.py`. While running `get_status()`, the `Query` also repeatedly attempts to retrieve params until found.
  - Streaming occurs in `tool.py` as part of `_wait_for_threads_to_finish`, once the params are retrieved.
- Refactors `_wait_for_threads_to_finish` into an async function that prints the output stream while simultaneously checking thread health.
  - Refactors how `elapsed_seconds` is passed around in `Tool`.
  - Refactors non-async usages of `_wait_for_threads_to_finish`.
- Adds an integration test for `python3` with streaming. The test also tests the printed stdout lines from the stream.
  - Adds streaming as part of the `toolchest.python3` call for the `python3_with_docker` test.

Known issues: 
- The `lug` package doesn't have an option to disable streaming (for remote execution) yet. (A PR for this has been created in [the lug repo](https://github.com/trytoolchest/lug/pull/24).)
- Streamed outputs from tool calls in `lug` may occur before the "Starting execution" message is printed by the `lug` script.
- ~~Docker pull messages that occur during command execution will be streamed (since they are printed to stdout).~~ (This was resolved by changes in other repos.)


tests/test_python3.py
```diff
@@ -63,3 +69,59 @@ def test_python3_with_docker():
...
+import time
+print("attempting to print timestamps every second")
+for _ in range(5):
+    message = datetime.datetime.utcnow().isoformat() + "Z"
+    print(message)
+    time.sleep(1)
+with open("./output/output.txt", "w") as f:
+    f.write("Success")
+            """
+        test_script_file.write(script)
```
lebovic, line 95: Could this be a normal file instead of a string that's written to a file?

tests/test_python3.py
```diff
@@ -63,3 +69,59 @@ def test_python3_with_docker():
...
+            number_of_expected_following_timestamps = 5
+        elif number_of_expected_following_timestamps > 0:
+            timestamp = datetime.datetime.fromisoformat(line.rstrip("Z"))
+            print(timestamp)
+            # Assert that each pair of timestamps occurred within 2 seconds
+            if last_timestamp:
+                time_diff = timestamp - last_timestamp
+                assert time_diff < datetime.timedelta(seconds=2)
+            last_timestamp = timestamp
+            number_of_expected_following_timestamps -= 1
```
lebovic, line 127: This is a bit complicated for "check that this output was streamed". Is there a way that you can just make this check against a constant?

toolchest_client/api/query.py
```diff
@@ -550,3 +569,23 @@ def get_job_status(self, return_error=False):
...
+            if server_setup_complete:
+                streaming_token = streaming_attributes.get("streaming_token")
+                streaming_ip_address = streaming_attributes.get("streaming_ip_address")
+                streaming_tls_cert = streaming_attributes.get("streaming_tls_cert")
+
+                self.streaming_client.initialize_params(
+                    streaming_token=streaming_token,
+                    streaming_ip_address=streaming_ip_address,
+                    streaming_tls_cert=streaming_tls_cert,
+                )
```
lebovic, line 591: I think it'd be cleaner to merge these two blocks, but it's on the border ¯\_(ツ)_/¯

toolchest_client/api/streaming.py
```diff
@@ -0,0 +1,53 @@
...
+
+    """
+
+    def __init__(self):
+        self.ssl_context = None
+        self.streaming_token = None
+        self.streaming_ip_address = None
+        self.streaming_tls_cert = None
+        self.username = "toolchest"
+        self.params_initialized = False
```
lebovic, line 28: `params_initialized` is a bit confusing of a name, is the meaning the same as `initialized`?

toolchest_client/api/streaming.py
```diff
@@ -0,0 +1,53 @@
...
+    async def receive_stream(self):
+        uri = f"wss://{self.username}:{self.streaming_token}@{self.streaming_ip_address}:8765"
+        async with websockets.connect(uri, ssl=self.ssl_context) as websocket:
+            stream_is_open = True
+            while stream_is_open:
+                try:
+                    stream_lines = await websocket.recv()
+                    print(stream_lines, end="")
+                except ConnectionClosed:
+                    stream_is_open = False
```
lebovic, line 50: Maybe a `print("Connection closed by server")` or something would be helpful here?

toolchest_client/tools/tool.py
```diff
@@ -322,25 +339,47 @@ def _check_thread_health(self):
...
+                else:
+                    self._pretty_print_pipeline_segment_status()
+
+                self.elapsed_seconds += increment_seconds
+                await asyncio.sleep(increment_seconds)
+
+            # Wait for the stream to complete before printing final status line
+            if self.streaming_task:
+                await self.streaming_task
+
```
lebovic, line 378: This is very nested, is there a cleaner way to do this?

toolchest_client/tools/tool.py
```diff
@@ -310,7 +319,15 @@ def _kill_query_threads(self):
             if thread.is_alive() and thread_status not in {ThreadStatus.COMPLETE, ThreadStatus.FAILED}:
                 self.query_thread_statuses[thread_name] = ThreadStatus.INTERRUPTING
 
-        self._wait_for_threads_to_finish(check_health=False, wait_for_upload=False)
+        # Wait for remaining threads to finish terminating
+        increment_seconds = 5
+        for thread in self.query_threads:
+            while thread.is_alive():
+                self._pretty_print_pipeline_segment_status()
+                self.elapsed_seconds += increment_seconds
```
lebovic, line 327: Why was `elapsed_seconds` moved to a class attribute instead of keeping it local to this function?

toolchest_client/api/streaming.py
```diff
@@ -0,0 +1,53 @@
...
+    Provides an interface to output lines streamed from the server.
+
+    """
+
+    def __init__(self):
+        self.ssl_context = None
+        self.streaming_token = None
+        self.streaming_ip_address = None
+        self.streaming_tls_cert = None
+        self.username = "toolchest"
```
lebovic, line 27: This is used once as a constant, I don't think it should be a class attribute

toolchest_client/api/query.py
```diff
@@ -44,7 +44,8 @@ class Query:
     # Max number of retries on status check timeouts.
     RETRY_STATUS_CHECK_LIMIT = 5
 
-    def __init__(self, stored_output=None, is_async=False, pipeline_segment_instance_id=None):
+    def __init__(self, stored_output=None, is_async=False, pipeline_segment_instance_id=None,
+                 streaming_enabled=False, stored_streaming_client=None):
```
lebovic, line 48: Why `stored_streaming_client` instead of the same name as the class attribute + name as it's passed as a parameter (`streaming_client`)?

toolchest_client/tools/tool.py
```diff
@@ -84,6 +86,11 @@ def __init__(self, tool_name, tool_version, tool_args,
         self.custom_docker_image_id = custom_docker_image_id
         self.instance_type = instance_type
         self.volume_size = volume_size
+        # auto-disable streaming if job is async
+        self.streaming_enabled = False if self.is_async else streaming_enabled
+        self.streaming_client = None
+        self.streaming_task = None
```
lebovic, line 92: What is the difference between `streaming_client` and `streaming_task`?

tests/test_python3.py
bcai2: It could, so I'll refactor it.

toolchest_client/tools/tool.py
bcai2: We also use `elapsed_seconds` in `_kill_query_threads`. (In fact, we were resetting `elapsed_seconds` to 0 before.)

toolchest_client/api/query.py
bcai2: No strong reason, I'll refactor

bcai2: Actually, it's to match the syntax of `stored_output`, but since `self.streaming_client` doesn't generate a new `StreamingClient` object if it isn't given, there's no reason to call it anything different, so I'll go ahead with the refactor to `streaming_client`.

toolchest_client/api/streaming.py
bcai2: Essentially, so I'll refactor it

toolchest_client/tools/tool.py
bcai2: `streaming_client` is the `StreamingClient` object, which contains the streaming parameters and websocket client.

`streaming_task` is an `asyncio` task that runs the `streaming_client.receive_stream()` function asynchronously as the status is checked. I can refactor this name to `streaming_asyncio_task` for clarity.

tests/test_python3.py
bcai2: Yup, by simplifying the script; refactored

toolchest_client/tools/tool.py
bcai2: Not by much, but I've removed some of the nesting.

toolchest_client/tools/tool.py
lebovic: Un-resolving this thread – `_kill_query_threads` is called from this function, could it be passed to `_kill_query_threads` as a function parameter instead of a class attribute?

lebovic: Ah I see, it's used several other places. Checks out

toolchest_client/tools/tool.py
```diff
@@ -322,25 +339,49 @@ def _check_thread_health(self):
...
                 raise ToolchestException("A job irrecoverably failed. See logs above for details.")
 
-    def _wait_for_threads_to_finish(self, check_health=True, wait_for_upload=True):
-        """Waits for all jobs and their corresponding threads to finish while printing their statuses."""
-        elapsed_seconds = 0
+    def _start_or_check_stream(self):
+        # Start streaming (in place of regular status updates).
+        if not self.streaming_asyncio_task:
+            # Clear the last status line before streaming begins
+            print("Printed lines:".ljust(120))
```
lebovic, line 346: Is all output streaming prefaced with `Printed lines:`?

If so, I don't think we should. It's the equivalent of using tmpenv with a banner at the top saying "remote environment".

toolchest_client/tools/tool.py
```diff
@@ -322,25 +339,49 @@ def _check_thread_health(self):
...
+            self.streaming_asyncio_task = asyncio.create_task(self.streaming_client.receive_stream())
+        # Check if an exception was raised in the streaming task
+        elif self.streaming_asyncio_task.done() and self.streaming_asyncio_task.exception():
+            self._kill_query_threads()
+            error_message = (
+                "A streaming error was encountered. See logs above for details.\n"
+                "Try re-running the same Toolchest command with this argument:\n"
+                "\tstreaming_enabled=False"
```
lebovic, line 354: Does this only happen with streaming errors, or on other worker node errors?

toolchest_client/tools/tool.py
```diff
@@ -497,6 +543,12 @@ def run(self):
                 "For support, contact Toolchest with the error log (above) and the following details:\n\n"
                 f"run_id: {pretty_print_run_ids}\n"
             )
+            if self.streaming_enabled:
+                print(
+                    "The error may have occurred during output streaming. "
+                    "Try re-running the same Toolchest command with this argument:\n"
+                    "\tstreaming_enabled=False"
```
lebovic, line 550: I think this message degrades the UX because:
- It's the last thing printed, and people read error messages bottom to top. If I saw this, I'd just retry with `streaming_enabled=False`.
- It's distracting. We should only print something in an error message / traceback if it's directly relevant to the error, this would be better in a "troubleshooting" section of the docs.

lebovic (changes requested):
Smaller UX comments, otherwise looks good!

toolchest_client/tools/tool.py
lebovic: More clear now! It looks like this is no PR comment is longer attached to the block of code, but responding here for continuity,

There's always a way to make things less nested: moving the bulk of the nested logic into its own function (e.g. how `_check_thread_health` works), which could make this more readable.

As-is, the function is good, but thought I'd mention that there is still a clear path toward removing nesting.

toolchest_client/tools/tool.py
lebovic: 👍 much clearer

toolchest_client/tools/tool.py
bcai2: This can happen with other worker node errors, but only if `streaming_enabled=True`.

lebovic (approved):

PR #270, opened 2022-10-10
feat: revamp docs

lebovic:


jherr-dev (approved):
Structure looks good. Skipped most content as that was ported from old docs.

bcai2 (approved):

PR #271, opened 2022-10-14
feat: add support for public images in dockerhub

jherr-dev:
This pr assumes if an image isn't found locally that it's a public image on dockerhub. A local image is still used if a matching one is found even if a public dockerhub image of the same name and tag exists. If no local or public image exists it will fail during execution instead of during the pre-execution client steps.

### DIFF
--- a/tests/test_python3.py
+++ b/tests/test_python3.py
@@ -63,3 +63,24 @@ def test_python3_with_docker():
     assert output_file.readline() == "[[ 58  64]\n"
     assert output_file.readline() == " [139 154]]"
     output_file.close()
+
+
+@pytest.mark.integration
+def test_python3_with_public_docker():
+    """
+    Tests using a public docker image with the write test script
+    """
+
+    test_dir = "./temp_test_python3/with_public_docker"
+    os.makedirs(f"{test_dir}", exist_ok=True)
+    toolchest.python3(
+        tool_args="./input/example.fastq",
+        script="s3://toolchest-integration-tests/write_test.py",
+        inputs="s3://toolchest-integration-tests/example.fastq",
+        output_path=f"{test_dir}/",
+        custom_docker_image_id="python:alpine3.16",
+    )
+
+    output_file = open(f"{test_dir}/output.txt", "r")
+    assert output_file.readline() == "Success"
+    output_file.close()
--- a/tests/test_python3.py
+++ b/tests/test_python3.py
@@ -74,7 +74,6 @@ def test_python3_with_public_docker():
     test_dir = "./temp_test_python3/with_public_docker"
     os.makedirs(f"{test_dir}", exist_ok=True)
     toolchest.python3(
-        tool_args="",
         script="s3://toolchest-integration-tests/write_path.py",
         output_path=f"{test_dir}/",
         custom_docker_image_id="python:alpine3.16",
--- a/tests/test_python3.py
+++ b/tests/test_python3.py
@@ -74,13 +74,14 @@ def test_python3_with_public_docker():
     test_dir = "./temp_test_python3/with_public_docker"
     os.makedirs(f"{test_dir}", exist_ok=True)
     toolchest.python3(
-        tool_args="./input/example.fastq",
-        script="s3://toolchest-integration-tests/write_test.py",
-        inputs="s3://toolchest-integration-tests/example.fastq",
+        tool_args="",
+        script="s3://toolchest-integration-tests/write_path.py",
         output_path=f"{test_dir}/",
         custom_docker_image_id="python:alpine3.16",
     )
 
     output_file = open(f"{test_dir}/output.txt", "r")
-    assert output_file.readline() == "Success"
+    assert output_file.readline() == "['/data/home/ec2-user/input', '/usr/local/lib/python310.zip', " \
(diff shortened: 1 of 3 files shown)


lebovic (approved):

tests/test_python3.py
```diff
@@ -7,9 +7,12 @@
 import toolchest_client as toolchest
 from toolchest_client.api.instance_type import InstanceType
 
-toolchest_api_key = os.environ.get("TOOLCHEST_API_KEY")
-if toolchest_api_key:
-    toolchest.set_key(toolchest_api_key)
+
+os.environ["BASE_URL"] = "https://staging.api.toolche.st"
+toolchest.set_key("ZjVhMmE.NDA1MTRhMzMtMTMyZC00YmU4LWE2NzEtZDFhYThiNzRiZGJj")
```
lebovic, line 12: !

tests/test_python3.py
```diff
@@ -63,3 +66,24 @@ def test_python3_with_docker():
...
+        tool_args="./input/example.fastq",
+        script="s3://toolchest-integration-tests/write_test.py",
+        inputs="s3://toolchest-integration-tests/example.fastq",
+        output_path=f"{test_dir}/",
+        custom_docker_image_id="python:alpine3.16",
+    )
+
+    output_file = open(f"{test_dir}/output.txt", "r")
+    assert output_file.readline() == "Success"
+    output_file.close()
```
lebovic, line 89: This tests that the command succeeds, but it doesn't test that it actually executed in the image that's specified nor that this works for an image with a registry URL.

I would be okay with merging this as-is because it it's a net improvement, but imo this isn't testing the feature.

toolchest_client/api/query.py
```diff
@@ -336,12 +335,12 @@ def _upload(self, input_file_paths, input_prefix_mapping, input_is_compressed):
...
+            return None
         try:  # Try to get the image before creating a repository.
             client = docker.from_env()
             image = client.images.get(custom_docker_image_id)
         except ImageNotFound:
-            raise ToolchestException(f"Unable to find image {custom_docker_image_id}.")
+            return custom_docker_image_id
         except (APIError, DockerException):
             raise EnvironmentError('Unable to connect to Docker. Make sure you have docker installed and that it is '
                                    'currently running.')
```
lebovic, line 346: If I'm reading this correctly, this error would still be hit if run somewhere that doesn't have Docker installed and running – is that right?

If so, this limits the upside of this feature; we would no longer require the image upload, but this wouldn't run somewhere like a hosted Jupyter notebook.

lebovic: Sorry for not catching this in the last review! I think it's worth fixing before merging.

lebovic (changes requested):

tests/test_python3.py
jherr-dev: It is testing the feature (although maybe not a clearly as a custom image). If the image doesn't have python3 command accessible or failed to download the tests fails. If the image exists locally it would upload which is a downside but that won't be the case currently.

lebovic: Yeah, I see that this test would fail if a Docker error is thrown (on client or on worker node) or if the image `python:alpine3.16` doesn't have `python3`. What I see missing as missing is that this also works on the base instance – nothing about this test is tied to the custom Docker image. I believe that it works now, but as a regression test I think it's suboptimal.

For example, with Lug, we had a very similar test for custom images (function runs and returns value X). However, the "function runs and returns value X" wasn't unique to the custom image; it succeeded even if the command wasn't rerouted to the custom image. That's why we didn't realize it wasn't functioning as intended for a while: the test passed, but if a component failed silently there was no test failure.

I see the same thing here: yes, this hits the path that a custom docker image upload path would hit. But this integration test isn't testing that:
- the Docker image uploaded
- the command is executing in that Docker image

> If the image exists locally it would upload which is a downside but that won't be the case currently.

I have that image locally already!

toolchest_client/api/query.py
```diff
@@ -342,8 +342,9 @@ def _upload_docker_image(self, custom_docker_image_id):
         except ImageNotFound:
             return custom_docker_image_id
         except (APIError, DockerException):
-            raise EnvironmentError('Unable to connect to Docker. Make sure you have docker installed and that it is '
-                                   'currently running.')
+            print('Unable to connect to Docker. Assuming image is accessible in a registry. If the image is hosted '
```
lebovic, line 345: > If the image is hosted locally start Docker and retry.

Nit: missing `,` :P

tests/test_python3.py
```diff
@@ -74,13 +74,14 @@ def test_python3_with_public_docker():
     test_dir = "./temp_test_python3/with_public_docker"
     os.makedirs(f"{test_dir}", exist_ok=True)
     toolchest.python3(
-        tool_args="./input/example.fastq",
-        script="s3://toolchest-integration-tests/write_test.py",
-        inputs="s3://toolchest-integration-tests/example.fastq",
+        tool_args="",
```
lebovic, line 77: Legit question: is `tool_args=""` needed?

lebovic (approved):

tests/test_python3.py
jherr-dev: No it's the default arg and can be removed

PR #274, opened 2022-10-24
fix: move TRANSFERRED_FROM_CLIENT status update

bcai2:
Fixes bug where the server could start and the server-side stream could time out before `_upload_docker_image` completed (e.g., when pushing a large docker image from the client).

### DIFF
--- a/toolchest_client/api/query.py
+++ b/toolchest_client/api/query.py
@@ -172,6 +172,7 @@ class Query:
         self._upload_docker_image(custom_docker_image_id)
         self._check_if_should_terminate()
 
+        self._update_status(Status.TRANSFERRED_FROM_CLIENT)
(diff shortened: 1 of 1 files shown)


toolchest_client/api/query.py
```diff
@@ -172,6 +172,7 @@ def run_query(self, tool_name, tool_version, input_prefix_mapping,
         self._upload_docker_image(custom_docker_image_id)
         self._check_if_should_terminate()
 
+        self._update_status(Status.TRANSFERRED_FROM_CLIENT)
```
jherr-dev, line 175: Nit: This might be better on L173 before the  `chech_if_should_terminate`, so the connection with the upload is more apparent.

jherr-dev (approved):

PR #281, opened 2022-10-27
feat: centrifuge base

bcai2:
- Adds the `centrifuge` tool.
- Adds docs.
- Refactors how `prefix_mapping` is generated for `megahit` with a new module (`input_util.py`) and function (`convert_input_params_to_prefix_mapping`). Adds a unit test for the function.


toolchest_client/tools/tool_args.py
```diff
@@ -206,6 +206,29 @@
...
+            "--al-conc",
+            "--un-gz",
+            "--un-bz2",
+            "--al-gz",
+            "--al-bz2",
+            "--un-conc-gz",
+            "--un-conc-bz2",
+            "--al-conc-gz",
+            "--al-conc-bz2",
+        ]
```
lebovic, line 230: Does this config actually reject non-blacklisted args that aren't the first item in `tool_args`?

toolchest_client/files/input_util.py
```diff
@@ -0,0 +1,57 @@
+"""
+toolchest_client.files.input_util
```
lebovic, line 2: I agree that this is a good point to separate this logic into it's own module.

The name (`input_util`) and logical grouping ("Utilities for handling input files/paths") doesn't make sense to me, though. What functions do we put in here, but not `general` or another module that pops up for input files or output files?

Would this fit in `general`, or could this be renamed?

lebovic (approved):

toolchest_client/files/input_util.py
bcai2: Done re: the unit test! It's in `toolchest_client/files/tests/test_input_params.py`.
I agree that the name isn't great. I'd like to refactor it, but I'm still trying to think of a logical name.

toolchest_client/tools/tool_args.py
bcai2: I'm not sure what you mean by this, can you give me an example config? (I don't believe so if I'm understanding correctly, but I just want to make sure that I am.)

toolchest_client/files/input_util.py
bcai2: I'm going to refactor this into `general.py` and move on.

bcai2:
@lebovic I'm merging this, but I also left replies and made a refactor that would be good to look over.

PR #288, opened 2022-11-29
feat: retain directory structure for compressed inputs

jherr-dev:
May be worth renaming `compress_inputs` to `retain_directories` or similar in the future. It's only really used for that now.

### DIFF
--- a/docs/docs/getting-started/using-files.md
+++ b/docs/docs/getting-started/using-files.md
@@ -17,9 +17,32 @@ Let's take a look at what it looks like to use different types of local and remo
 Local files are the most intuitive: you just pass normal paths directly to Toolchest. In the background, the files are 
 transferred to and from the cloud.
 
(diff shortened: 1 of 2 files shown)


docs/docs/getting-started/using-files.md
```diff
@@ -17,9 +17,32 @@ Let's take a look at what it looks like to use different types of local and remo
 Local files are the most intuitive: you just pass normal paths directly to Toolchest. In the background, the files are 
 transferred to and from the cloud.
 
-`inputs` takes files or directories. If a directory is passed, all files within the directory are used as input.
-
-`output_path` takes a directory. Output files are written in this directory.
+`inputs` takes file and/or directory path(s).
```
lebovic, line 20: `and/or` and `(s)` is hard to parse, I'd keep as something simpler like "`inputs` takes paths to files for directories"

lebovic (approved):
👍

PR #298, opened 2022-12-12
fix: remove extra threading layer

lebovic:
Generally maintains the pipeline segment instance vs. task separation, but removes the unnecessary threading overhead. Also removes some client-size outdated parallelization logic, and refactors `thread_status` -> `pretty_status` to re-use much of the same internal logic.

Tested to work with basic Python output streaming + the test command, but high risk for thorough integration test failures. Not yet tested with Jupyter notebooks or pipeline managers.


toolchest_client/api/query.py
```diff
@@ -38,55 +39,53 @@ class Query:
     """
 
     # Period (in seconds) between requests when waiting for job(s) to finish executing.
-    WAIT_FOR_JOB_DELAY = 1
-    # Multiple of seconds used when pretty printing job status to output.
-    PRINTED_TIME_INTERVAL = 5
+    WAIT_FOR_JOB_DELAY = 5
```
lebovic, line 42: Updated the delay to 5 seconds for the API call, as we were already waiting 5 seconds for the printed interval. This leads to someone off-kilter status updates, as they're not updated at the five second mark.

toolchest_client/api/status.py
```diff
@@ -43,7 +43,7 @@ class Status(str, Enum):
     FAILED = "failed"
 
 
-class ThreadStatus(str, Enum):
+class PrettyStatus(str, Enum):
```
jherr-dev, line 46: Any reason this is `PrettyStatus` and not just `Status`. I understand the pretty print function having that name but the status to me can just be `Status`.

jherr-dev (changes requested):
1 issue and 1 style nit.

toolchest_client/api/status.py
lebovic: The `Status` enum above `PrettyStatus` corresponds 1:1 with API-side statuses. The `PrettyStatus` enum (formerly `ThreadStatus` was previously for tracking thread statuses.

There wasn't much logic tied to the `PrettyStatus/ThreadStatus` except for pretty-printing status (e.g. just `executing`, no `awaiting_execution` / `beginning_execution` / etc).

I think that PrettyStatus should be removed – or exist as a mapping from Status to a printable status – but I kept it in place to avoid making too extensive of a refactor. Open to removing it in this PR, though.

toolchest_client/api/output.py
```diff
@@ -75,10 +76,14 @@ def download(self, output_path=None, output_dir=None, skip_decompression=False):
...
+
     def get_status(self, **kwargs):
         """
-        Returns the status of a run. Only for use when the Output instance is initialized with a run_id.
+        Returns the status of a run
         """
         if not self.run_id:
             raise ValueError("Cannot get status on an output that has no run_id")
-        return get_status(self.run_id, **kwargs)
+        self._refresh_status(**kwargs)
```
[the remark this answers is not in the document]
lebovic: Huh, it's a miracle that didn't break anything: there's an imported `get_status` function that hits the API, _and_ an internal function for getting status.

Refactored the import and `refresh_status` call!

toolchest_client/api/status.py
jherr-dev: Ah didn't expand the file and see `Status` above. I would be okay leaving the refactor for later especially since multi-value enums in python aren't great.

jherr-dev (approved):
LGTM

PR #300, opened 2023-01-19
v0.11.9

lebovic:


toolchest_client/api/output.py
```diff
@@ -75,10 +76,14 @@ def download(self, output_path=None, output_dir=None, skip_decompression=False):
         )
         return self.output_file_paths
 
+    def refresh_status(self, **kwargs):
```
jherr-dev, line 79: Why make this an object value?

jherr-dev (approved):
LGTM but have a style question

PR #301, opened 2023-01-23
feat: add log levels

lebovic:
This also excludes https://www.flake8rules.com/rules/E731.html from our lint check; I don't think it's worth a debate if you disagree, so I'm happy to roll back if needed.


.flake8
```diff
@@ -3,6 +3,8 @@ max-line-length = 120
 exclude=
     demo.py
     docs/conf.py
+ignore =
+    E731
```
jherr-dev, line 7: Don't love global ignores. [This rule](https://www.flake8rules.com/rules/E731.html) does makes sense to me as well despite the worse code style around lambdas.

toolchest_client/files/s3.py
```diff
@@ -114,6 +116,24 @@ def pretty_print_file_size(num_bytes):
     return f"{(num_bytes / (1024 ** abbreviation_index)):.1f}{pretty_abbreviation[abbreviation_index]}"
 
 
+def pretty_status_print(filename, seen_so_far, size, percentage):
+    # This doesn't format well as a progress bar with the logger, but we want to hide unless at DEBUG level
+    # print("log level", get_log_level())
+    if get_log_level() == "DEBUG":
```
jherr-dev, line 122: Optionally move condition out a level.

toolchest_client/tools/api.py
```diff
@@ -48,13 +49,13 @@ def alphafold(inputs, output_path=None, model_preset=None, max_template_date=Non
     """
     if 'instance_type' in kwargs:
         raise ToolchestException("Argument 'instance_type' is not supported by Alphafold currently.")
-    tool_args = (
-        (f"--model_preset={model_preset} " if model_preset is not None else "") +
-        (f"--max_template_date={max_template_date} " if max_template_date is not None
-         else f"--max_template_date={date.today().strftime('%Y-%m-%d')} ") +
-        (f"--is_prokaryote_list={is_prokaryote_list} " if is_prokaryote_list is not None else "") +
-        ("--db_preset=reduced_dbs " if use_reduced_dbs else "")
-    )
```
jherr-dev, line 57: I personally think this is preferable to the reformatted code without the wrapping ()

jherr-dev (approved):
LGTM. Have some comments around lint and style

PR #302, opened 2023-01-23
v0.11.10

lebovic:


tests/test_clustalo.py
```diff
@@ -10,7 +10,8 @@
 
 
 @pytest.mark.integration
-def test_clustalo_standard():
```
lebovic, line 13: Worth refactoring `AWS` / `on-prem` before merge

lebovic: Refactoring

lebovic: Refactored

PR #308, opened 2023-02-17
feat: improve run output

jherr-dev:
Fixes the issues from the ticket. Docker in progress still could be improved but should give some indication and updates to a final "uploaded" message.

### DIFF
--- a/toolchest_client/api/query.py
+++ b/toolchest_client/api/query.py
@@ -40,7 +40,7 @@ class Query:
     """
 
     # Period (in seconds) between requests when waiting for job(s) to finish executing.
-    WAIT_FOR_JOB_DELAY = 5
+    WAIT_FOR_JOB_DELAY = 1
     # Max number of retries on status check timeouts.
     RETRY_STATUS_CHECK_LIMIT = 5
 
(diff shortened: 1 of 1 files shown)


toolchest_client/api/query.py
```diff
@@ -404,17 +404,19 @@ def _upload_docker_image(self, custom_docker_image_id):
...
-                )
-                sys.stdout.flush()
+                if counter % 10 == 0:
+                    dots = '.' * (((counter // 10) % 5) + 1)
+                    sys.stdout.write(
+                        f"\rDocker image uploading{dots}",
+                    )
+                    sys.stdout.flush()
+                counter += 1
+            print("\rDocker image uploaded")
```
lebovic, line 419: This should be a `logger.info`

toolchest_client/api/query.py
```diff
@@ -404,17 +404,19 @@ def _upload_docker_image(self, custom_docker_image_id):
...
                 # Print doesn't work on some consoles, like the JetBrains suite
-                sys.stdout.write(
-                    "\r{} {}".format(
-                        update.get("status", ""),
-                        update.get("progress", ""),
-                    ).ljust(120),
-                )
-                sys.stdout.flush()
+                if counter % 10 == 0:
+                    dots = '.' * (((counter // 10) % 5) + 1)
```
lebovic, line 413: Maybe worth showing the same style of progress indicator here as below?

lebovic (approved):
LGTM else nit

PR #314, opened 2023-03-22
feat: enable volume persistence with volume name or universal function

jherr-dev:


toolchest_client/api/query.py
```diff
@@ -111,6 +112,9 @@ def run_query(self, tool_name, tool_version, input_prefix_mapping,
         :param skip_decompression: Whether to skip decompression of the output file, if it is an archive.
         :param instance_type: instance type that the tool will run on.
         :param volume_size: size of the volume for the instance.
+        :param universal_volume_name: the name of the volume that should be attached to the current run or persisted
+            after.
+        :param universal_name: the name of the universal function to pull parameters from.
```
lebovic, line 117: Nit: "name of universal volume" and "name of universal function" would suffice here

toolchest_client/api/query.py
```diff
@@ -229,7 +238,11 @@ def _send_initial_request(self, tool_name, tool_version, tool_args, database_nam
         try:
             create_response.raise_for_status()
         except HTTPError:
-            logger.error("Query creation failed.", file=sys.stderr)
+            error_message = "."
```
lebovic, line 241: I see how `"."` fits as the default value here, but reading this is a bit confusing – I'd opt for keeping this as `""` and not having a final period if there's a 4xx error

toolchest_client/tools/api.py
```diff
@@ -805,7 +805,7 @@ def lastal5(output_path=None, output_primary_name="out.maf", inputs=[], database
 
 def lug(script, tool_version, custom_docker_image_id, container_name, docker_shell_location, inputs=None,
         output_path=None, tool_args="", instance_type=InstanceType.COMPUTE_2, volume_size=8, streaming_enabled=True,
-        pip_dependencies=None, **kwargs):
+        pip_dependencies=None, universal_name=None, **kwargs):
```
lebovic, line 808: Why is universal name exposed here, but not universal volume name? Is either needed, or can they be passed as kwargs?

toolchest_client/tools/lug.py
```diff
@@ -13,7 +13,8 @@ class Lug(Tool):
     """
     The Lug implementation of the Tool class.
     """
-    def __init__(self, tool_args, inputs, output_path, tool_version, custom_docker_image_id=None, **kwargs):
+    def __init__(self, tool_args, inputs, output_path, tool_version,
+                 custom_docker_image_id=None, universal_name=None, **kwargs):
```
lebovic, line 17: Same question here as above for universal name vs. volume inclusion

lebovic (approved):
LGTM, you may want to change how "universal_x" are defined in the Lug function params

